PR #4801, opened 2023-12-30
WIP: Branch/rutf8 simd

gitlab-importer:
In GitLab by @planrich on Feb 28, 2020, 13:21

This helps to speed up the calls to count the utf8 code points (e.g. len(unicodestr)). On the benchmark on python2 this is pretty insignificant. Doing a lot of string length calls in a loop on unicode objects shows speedup. The effect would be much more interesting on Python 3.X


mattip:
Also, see [this post](https://cceckman.com/writing/branchless-utf8-encoding/) about branchless encoding of 32-bit unicode to utf-8

PR #4802, opened 2023-12-30
WIP: Get rid of assert reinterpretation in rpython tests

gitlab-importer:
In GitLab by @rlamy on Jul 22, 2020, 21:44

Assert reinterpretation has long been deprecated in pytest and was removed in version 3.0. This branch allows rpython tests to run with assert rewriting by adding `rpython.rlib.objectmodel.assert_()`, which behaves like an RPython `assert` statement but doesn't get rewritten by pytest.

Fixing the test suite is then just a matter of mechanically replacing all asserts in RPython functions defined in test files by this new `assert_()` function.


mattip:
Losing. Please reopen and fix conflicts if it is worth redoing.

PR #4806, opened 2023-12-30
mp queue: fix close without any put

### LINKED ISSUE #3372: multiprocessing.queues.Queue leaks pipes under certain conditions (CPython issue42752)
In Heptapod by @crazycasta on Jan 1, 2021, 03:04

In writing a test for #3365 I noticed there's a bit of a problem with multiprocessor.queues.Queue. The trouble is if you don't actually put anything in the Queue (say for instance you're only using it to receive in the main process) the close method won't actually lead to the write pipe closing. Since the Queue is buffer and uses a thread, the closing of the write pipe is not exactly trivial. Branch py3_7_mp_queue_close_fix is my suggested fix.

gitlab-importer:
In GitLab by @crazycasta on Jan 1, 2021, 05:14

This fixes #3372.

In short:
* Current CPython multiprocessing.queues.Queue doesn't close properly unless put has been called before you try to close.
* Not as simple as just changing the write pipe close as this has to be done in the context of the buffer thread.
* This makes sure that before the thread has been started (before put has been called for instance) that something will close the write pipe.

### DIFF
--- a/lib-python/3/multiprocessing/queues.py
+++ b/lib-python/3/multiprocessing/queues.py
@@ -29,6 +29,7 @@ _ForkingPickler = context.reduction.ForkingPickler
 from .util import debug, info, Finalize, register_after_fork, is_exiting
 
 
+# PyPy added to fix issue bpo 42752
 class CleanExchange:
     def __init__(self, obj, attr):
         self.obj = obj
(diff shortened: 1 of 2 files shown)


mattip:
The test added here still fails on CPython3.13. Over at python/cpython#86918, if I understand correctly, there was some concern about triggering the failure in python/cpython#109370.

PR #4807, opened 2023-12-30
Python3 changes for rpython

gitlab-importer:
In GitLab by @mattip on Oct 13, 2021, 20:40

I had been working on this branch to update some of the rpython syntax to be python3-compatible. This is the first set of changes.


TheShermanTanker:
Sorry to interject, but does this mean that rpython will be written in Python 3 once merged, so translation can be done with a Python 3 Interpreter instead of Python 2?

olliemath:
> Sorry to interject, but does this mean that rpython will be written in Python 3 once merged, so translation can be done with a Python 3 Interpreter instead of Python 2?

Unfortunately no - there is a long way to go to support something like that: https://github.com/pypy/pypy/wiki/RPython-3-considerations

Any changes to rpython would need to support translating under both py2 and py3 because pypy is written in python2

auvipy:
May I ask, why python2 support is still very important? also if you allow outside contributors who want to contribute using these old pr?

auvipy:
Thanks, i will follow the wiki and links and come with an easily reviewable PR.

cfbolz:
> Note that since translation is tightly tied to understanding the AST, we will probably never be able to update freely from one version of Python3 to another, rather RPython3 will refer to a specific version of Python3.x.

it's actually worse than that: RPython is tied to a specific version of the Python *bytecode*, which is changing a lot in CPython 3.x versions due to the performance work of the faster CPython team. Therefore, Python 3.x is really a very tough-to-use technological basis for RPython at this point. It's not completely impossible but it would very hard work, that is not really plausibly done without serious sources of money.

maybe we should add an FAQ entry about this. to be quite frank, I don't see RPython3 happening.

rlamy:
@cfbolz I think it should be easier to deal with bytecode changes now than in the 2.x era. Indeed, numba seem to be keeping up fine. I feel that CPython core devs do consider bytecode disassembly as a legitimate use case nowadays - they seem to be paying *some* attention to it when making changes.

auvipy:
thanks for more context. I will be coming with a PR/Multiple PRs to improve python3 compatibility codes for rpython and other parts of PyPy. I hope to learn gradually in the process....

PR #4811, opened 2023-12-30
(Do not merge yet) Add simple float.max specialization

gitlab-importer:
In GitLab by @tekknolagi on Sep 28, 2023, 19:06

--HG--
branch : mb-dup

### DIFF
--- a/rpython/jit/metainterp/test/test_float.py
+++ b/rpython/jit/metainterp/test/test_float.py
@@ -84,6 +84,18 @@ class FloatTests:
         res = self.interp_operations(g, [-12345])
         assert type(res) is float and res == float(long(r_uint(-12345)))
 
+    def test_max(self):
+        def f(a, b):
+            return max(a, b)
+        res = self.interp_operations(f, [-5.25, 3])
+        assert res == 3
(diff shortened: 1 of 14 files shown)


rpython/jit/metainterp/test/test_float.py
```diff
@@ -84,6 +84,18 @@ def g(i):
         res = self.interp_operations(g, [-12345])
         assert type(res) is float and res == float(long(r_uint(-12345)))
 
+    def test_max(self):
+        def f(a, b):
+            return max(a, b)
+        res = self.interp_operations(f, [-5.25, 3])
+        assert res == 3
+        # TODO(emacs): Figure out why nan causes issues here
```
devdanzin, line 92: Would it be because "If only one value is a NaN (SNaN or QNaN) for this instruction, the second source operand, either a NaN or a valid floating-point value, is written to the result"?

https://www.felixcloutier.com/x86/maxsd

mattip:
@tekknolagi, @cfbolz: should we move forward with this?

tekknolagi:
I think we can drop it. I never tracked down the differences and don't have the time/energy to now :/

PR #4817, opened 2024-01-01
Give the named branches token write access to push.

Julian:
And limit how much we fetch so we don't wait ages.

### DIFF
--- a/.github/workflows/named-branches.yml
+++ b/.github/workflows/named-branches.yml
@@ -1,12 +1,17 @@
 name: 'Named Branches'
 on: push
+permissions:
+  contents: write
 jobs:
   named-branches:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mattip:
Thanks

mattip:
... and it is green. Maybe update the named-branches README to reflect this? Thanks for the quick response.

Julian:
> Maybe update the named-branches README to reflect this? 
Just did!

And my pleasure, does seem to work at least for a few of the recent commits! Hooray.

Julian:
Ah hrm wait, not sure yet actually, it reports 8e7f30b0a87 as being made on main, but that shouldn't be the case I think... Going to keep an eye out, I think maybe there is/was a change in GHActions having to do with detached head that might be affecting things.

PR #4818, opened 2024-01-01
fix limited API guard for METH_FASTCALL to match CPython

### LINKED ISSUE #4050: cpyext METH_FASTCALL and limited API
In Heptapod by @karlotness on Dec 28, 2023, 04:52

It looks like there is an issue with the limited API guard around `METH_FASTCALL` in `methodobject.h` which should be part of the limited API in Python 3.10.

The [current PyPy version](https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.10/pypy/module/cpyext/include/methodobject.h?ref_type=heads#L30) requires `0x03100000` while the [version in CPython 3.10](https://github.com/python/cpython/blob/b6535ea7ecf40c51a9a13089d961c29abeaffa5e/Include/methodobject.h#L82) uses `0x030A0000`. It looks like that might have been a CPython mistake that got imported into PyPy since it was fixed in the linked commit.

Very minor problem but it would be nice to have the inconsistency fixed since does cause some build issues.

karlotness:
This should fix `METH_FASTCALL` so it is included in the 3.10 limited API like in CPython.

Fixes #4050

### DIFF
--- a/pypy/module/cpyext/include/methodobject.h
+++ b/pypy/module/cpyext/include/methodobject.h
@@ -27,7 +27,7 @@ extern "C" {
 
 #define METH_COEXIST   0x0040
 
-#if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x03100000
+#if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x030A0000
 #define METH_FASTCALL  0x0080
 #endif
 


mattip:
Thanks @karlotness. PyPy does not support the limited API, but no reason to leave a typo.

karlotness:
Thanks!
Noted that the limited API isn't supported, but I do appreciate fixing this up. It helps reduce the effort to build against PyPy (since there's no need to special-case the macros in the extension) even if there's still a need to build against multiple versions. So thanks again for taking a look.

Issue #4816, opened 2024-01-01
Possible ABI breakage in 7.3.14


mgorny:
It seems that some ABI change was introduced between 7.3.13 and 7.3.14, more specifically between 2023-11-29 and 2023-12-07 snapshots. The Pillow extensions built with the prior version crash when used with pikepdf built with the latter version.

To reproduce:

```
git clone --depth 1 https://github.com/pikepdf/pikepdf/
cd pikepdf
pypy3 -m venv .venv
. .venv/bin/activate
# note that this installs pillow from official wheel
# also note that you need to have qpdf installed on the system
pip install .[test]
python -m pytest -o addopts= -s tests/test_image_access.py::test_image_save_compare
```

The result is:
```
========================================================= test session starts =========================================================
platform linux -- Python 3.10.13[pypy-7.3.14-final], pytest-7.4.4, pluggy-1.3.0
rootdir: /tmp/pikepdf
configfile: pyproject.toml
plugins: hypothesis-6.92.2, xdist-3.5.0, timeout-2.2.0, cov-4.1.0
collected 1 item                                                                                                                      

tests/test_image_access.py Fatal Python error: Segmentation fault

Stack (most recent call first, approximate line numbers):
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/PIL/Image.py", line 2980 in frombuffer
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/pikepdf/models/_transcoding.py", line 106 in image_from_byte_buffer
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/pikepdf/models/image.py", line 552 in _extract_transcoded_1248bits
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/pikepdf/models/image.py", line 605 in _extract_transcoded
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/pikepdf/models/image.py", line 635 in _extract_to_stream
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/pikepdf/models/image.py", line 672 in extract_to
  File "/tmp/pikepdf/tests/test_image_access.py", line 301 in test_image_save_compare
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/hypothesis/core.py", line 848 in run
  File "/tmp/pikepdf/.venv/lib/pypy3.10/site-packages/hypothesis/core.py", line 728 in default_executor
[...]

PR #4821, opened 2024-01-02
try to add a simple github action

cfbolz:


mattip:
cool. Of course this is not really portable since it builds on `ubuntu-20.04`. Is it worth building inside a `manylinux_2014` docker? The artifact does have value for bisecting failures like the one in #4816.

.github/workflows/translate.yml
```diff
@@ -0,0 +1,53 @@
+name: Translate PyPy
+
+on:
+  push:
+    branches: [ "main", "py3.9", "py3.10"]
+  pull_request:
+    branches: [ "main", "py3.9", "py3.10"]
```
mattip, line 7: This [will add](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) the ability to kick of this action via a button on the workflow page.

```suggestion
    branches: [ "main", "py3.9", "py3.10"]
  workflow_dispatch:    
```

cfbolz:
@mattip yeah, it's not properly portable at all, but without `--make-portable` it doesn't even work on my ubuntu 23.10 laptop, so this is still an improvement. I would like to merge this as it is, we can find out how to do Docker in a github action as a later step. ok for you?

mattip:
Yes, makes sense.

cfbolz:
cool, merging then :blush:

PR #4823, opened 2024-01-02
PR to check whether the CI run would work on py3.9 too

cfbolz:
shouldn't be merged via a PR, but I want to run the CI and see whether it also works on py3.9


pypy/tool/release/repackage.sh
```diff
@@ -1,11 +1,11 @@
 #! /bin/bash
 
 # Edit these appropriately before running this script
-pmaj=3  # python main version: 2 or 3
-pmin=9  # python minor version
+pmaj=2  # python main version: 2 or 3
+pmin=7  # python minor version
```
cfbolz, line 5: @mattip is this correct to merge to pypy3.9? or should the change be removed in the merge commit?

mattip: It doesn't matter. I fix up `pmaj`, `pmin` for each repackaging run, so the values that end up committed here are kind of random

Issue #3678, opened 2022-02-15
Add missing docstrings to all builtins


gitlab-importer:
In Heptapod by @mattip on Feb 15, 2022, 12:41

This script
```
import builtins
import doctests
tests = doctest.DocTestFinder().find(builtins)
print(len(tests)
```

prints 447 on PyPy and 810 on CPython. A newcomer task could be to copy the docstrings from CPython into PyPy. Since some of them are quite large, they should be separate from the code to keep the functions readable.

learning-dev:
@mattip, can I take this up if it is available?

mattip:
Sure. Please carefully consider how to do this without cluttering up the code.

learning-dev:
Hi @mattip, I have to add "doc strings" for the built-ins mentioned in this issue and not in the inked issue. I see that there already a PR raised for the linked issue.

mattip:
PR #4824 targeted `main`. Any new work should target `py3.10` or `py3.11` since those are the currently supported python3 branches. The technique used in #4824 is sound and can be adapted to future work. We don't assign issues, please reference this issue in your PR so people know you are working on it.

mattip:
Of course feel free to ask questions if something is unclear.

PR #4824, opened 2024-01-03
Add docstrings to tupleobject

3tty0n:
This PR is related to #3678.

First I tried add doctrings to tuple. Those doctriings I added are copied from Python 2.7.  

Should I add those from Python 3 instead of Python 2?

### DIFF
--- a/pypy/objspace/std/objectobject.py
+++ b/pypy/objspace/std/objectobject.py
@@ -221,42 +221,6 @@ def descr___format__(space, w_obj, w_format_spec):
         space.warn(space.newtext(msg), space.w_PendingDeprecationWarning)
     return space.format(w_as_str, w_format_spec)
 
-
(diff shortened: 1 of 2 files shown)


mattip:
If you target `main`, you should use the 2.7 docstings. If you target `py3.9`, you should use the python3 docstrings. 

Some of these like should be assigned to `object`, not `tuple` (like `object.__new__`).

3tty0n:
Thanks for reviewing!

> If you target main, you should use the 2.7 docstings. If you target py3.9, you should use the python3 docstrings.

Sure. I'll do it based on Python 3's docstrings when I target py3.9.

> Some of these like should be assigned to `object`, not `tuple` (like `object.__new__`).

 I moved some of them that should have been added to `object` and used them in `tuple.`

mattip:
Close/reopen to trigger CI

PR #4825, opened 2024-01-03
change PyBUF_MAX_NDIM back to 36 for ABI consistancy (issue #4816)

### LINKED ISSUE #4186: Fix handling of leading zeroes in integer literals. - [closed]
In Heptapod by bitbucket_importer on Jul 30, 2018, 16:03

_Merges bitbucket/declined-pr-620 -> branch/py3.5_

*Created originally on Bitbucket by [nurbldoff (Johan Forsberg)](https://bitbucket.org/[7.3.7](https://github.com/gitlab-importer/pypy/milestone/7)B34d9aa79-b3e9-49b0-a40c-ee18a99cb160%7D/)*

*Changesets for this Pull Request have not been imported, because it had been already declined on Bitbucket. Marked as closed by the import user.*

\(I started work on this fix during the Europython sprint.\)

This fixes the case of 

    int(“07”, base=0) 

which should raise a ValueError in python 3.

This obscure behavior seems to have been wrong in pypy for some time, but there was apparently no test for it until it broke some unrelated cpython tests in 3.6. I added some more checks to the test just to be sure.

See [https://docs.python.org/3/reference/lexical\_analysis.html#integer-literals](https://docs.python.org/3/reference/lexical_analysis.html#integer-literals) and PEP 3127 for more information.

mattip:
@mgorny, this is enough to fix #4186, right?

### DIFF
--- a/pypy/module/cpyext/parse/cpyext_object.h
+++ b/pypy/module/cpyext/parse/cpyext_object.h
@@ -73,8 +73,8 @@ typedef Py_ssize_t (*segcountproc)(PyObject *, Py_ssize_t *);
 typedef Py_ssize_t (*charbufferproc)(PyObject *, Py_ssize_t, char **);
 
 /* Py3k buffer interface, adapted for PyPy */
-/* XXX remove this constant, us a PyObject_VAR_HEAD instead */
-#define PyBUF_MAX_NDIM 64
+/* Changing this constant changes the ABI */
+#define PyBUF_MAX_NDIM 36
 typedef struct bufferinfo {
     void *buf;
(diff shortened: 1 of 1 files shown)


mgorny (approved):
Yep, this is what I've tested.

mgorny:
Do you happen to know if this change requires translating PyPy again, or is it sufficient to reinstall the headers?

mattip:
Since we parse and use bufferinfo internally in cpyext, this requires translating again. I will be releasing a 7.3.15 bugfix in a week or so, let's see if there are more bugs reported with 7.3.14.

PR #4827, opened 2024-01-04
Add action for rpython unit tests

olliemath:
I attempted to get these working a while back - hopefully this is helpful (let me know if you don't want to run these tests in this repository).

This adds the rpython unit tests and configures them to run only when any of `rpython/`, `.github/` or `pytest.py` change. Due to the lengthy tests I plan to parallelise (to decrease feedback times) - but possibly in a separate MR.

Here's the currently failing tests:

```
FAIL rpython/rlib/test/test_rdynload.py::TestDLOperations::()::test_ldscripts
FAIL rpython/rlib/test/test_rfile.py::TestFile::()::test_open_buffering_full
FAIL rpython/rlib/test/test_rfile.py::TestFile::()::test_fdopen_buffering_full
FAIL rpython/rlib/test/test_rgc.py::test_get_memory_usage
FAIL rpython/rlib/test/test_rposix_stat.py::test_high_precision_stat_time
FAIL rpython/rtyper/test/test_rlist.py::TestRlist::test_memoryerror 
/home/runner/work/_temp/55ca308d-8a8a-45fa-bc5b-9ae580713ba0.sh: line 1:  2000 Segmentation fault      (core dumped) pypy2 pytest.py rpython/rtyper -v
```

These are possibly due to running the tests under pypy2 rather than under cpython2 - but since we use pypy2 to translate pypy anyway, it seems like that should be the "happy path" and the tests need fixing


mattip:
> These are possibly due to running the tests under pypy2 

FWIW, the windows buildbot has been using pypy2 as a host python for years and does not see segfaults. Although there are a number of [failing rpython tests](https://buildbot.pypy.org/summary?builder=builder=rpython-win-x86-64). I think it is fine to add these as long as the time does not exceed, say, 20-25 minutes

mattip:
It would be nice if we could have a text file of tests to skip on CI, that might also come in handy for the flaky objspace CI run

olliemath:
> It would be nice if we could have a text file of tests to skip on CI, that might also come in handy for the flaky objspace CI run

ah you mean a list of flaky tests? yeah that's probably a good approach to start with

generally it seems like some tests are dependent upon interpreter implementation and os to pass, so there's either a genuine bug we should fix, or a `@skipif` we should add

olliemath:
have fixed most things - some of the failures here were also seen under the windows buildbot run so those should pass now too - the segfault hasn't re-ocurred since I fixed the other tests :shrug:

olliemath:
the only remaining issue is that the jit tests never complete - tempted to skip them for now - happens locally too and I suspect it's to do with flaky subprocess communication

these would pass under cpython2, but the "setup python" action does not have cpython2 as an option

mattip:
> the jit tests never complete

Does this also timeout on the windows buildbot?

rpython/rlib/test/test_rfile.py
```diff
@@ -135,11 +137,19 @@ def test_open_buffering_full(self):
...
+                # causing the second read bellow to fail without this:
+                f.write('z')
+
+            # This read should get all of the flushed data, but not everything
+            # (because our writes are buffered)
             assert f2.read() != ''
             f.close()
+            # Closing the file should flush the unread bytes
             assert f2.read() != ''
             f2.close()
```
mattip, line 154: Can these be rewritten to use a context manager instead of managing the open/close?

rpython/rtyper/lltypesystem/ll2ctypes.py
```diff
@@ -1087,8 +1087,7 @@ def ctypes2lltype(T, cobj, force_real_ctypes_function=False):
             except (ValueError, OverflowError):
                 for tc in 'HIL':
                     if array(tc).itemsize == array('u').itemsize:
-                        import struct
-                        cobj &= 256 ** struct.calcsize(tc) - 1
+                        cobj &= sys.maxunicode
```
mattip, line 1090: heh, nice

.github/workflows/rpython-unit-tets.yml
```diff
@@ -0,0 +1,174 @@
...
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "pypy2.7-7.3.10"
+    - run: |
+          echo "Install requirements"
+          sudo apt-get install -y libgc-dev
+          pypy2 -m pip install --user hypothesis==4.39.3
+    - name: Test Jit - Other
+      run: pypy2 pytest.py -vs rpython/jit --ignore=rpython/jit/backend/x86
```
mattip, line 98: This is going to be quite slow single-threaded. How long does it take to run locally?

olliemath: Here's a fully passing run where I replace pypy with cpython for the jit tests: https://github.com/olliemath/pypy-official/actions/runs/7420947044/job/20193310402

The timings are:

misc (pypy): 1 minute
**jit-x86 (cpython): 47 minutes**
jit-other (cpython): 18 minutes
memory (pypy): 10 minutes
rlib (pypy): 22 minutes
rtyper (pypy): 8 minutes
translate (pypy): 28 minutes

Under pypy those same jit tests are timing out after 6 hours (see here https://github.com/olliemath/pypy-official/actions/runs/7417566997/job/20184243634 ). Doesn't seem to happen on windows so I assume it's something specific to linux+pypy

The runners have 2 cores so could definitely speed that up

olliemath: Locally I can reproduce the timeout - at the end of some tests `rpython/tool/runsubprocess.py` occasionally hangs waiting for input from stdin

rpython/rlib/test/test_rfile.py
olliemath: I don't think we can manage both in a context manager, since context managers are "LIFO":
```python
with open(...) as f1:
  with open(...) as f2:
    # here we'd like to close f1, but can't until after f2 is closed
    # the last-most open is closed first
```

I could add a single `with`, but mixing paradigms might not be the best for readability

olliemath: I take that back, we can manually `f1.close()` before exiting the context

olliemath:
Seems the only remaining thing is the occasional segfaults in rtyper

olliemath:
Hopefully splitting out the offending test file makes it a bit more stable (rather than running everything in a single pytest session)

cfbolz:
ah, that test_rlist.py crash actually shows a bug in pypy2's range implementation, I think. I'll look into it.

cfbolz:
I'd say we merge this and then iterate. ok, @olliemath?

olliemath:
Sounds good to me :+1:

cfbolz:
done. thanks a ton for working on this!

PR #4828, opened 2024-01-04
Init sys.executable from GetModuleFileName on win32

### LINKED ISSUE #4003: pypy.exe -m ensurepip does not work when invoked from inside Windows Services for Linux
In Heptapod by @tekknolagi on Sep 24, 2023, 20:41

(posted on behalf of Andi McClure)

I use WSL (Windows Services for Linux) as my command line shell on Windows 10. I use basic tools such as grep, git etc inside WSL; for coding tools such as Rust cargo, Python, etc I usually (since WSL has some limitations) install Windows versions and invoke the Windows .exes from the Linux command line, which works fine, I just have to manually type the .exe at the end when I invoke them. In this respect I basically use WSL as if it were MSYS.
Today I decided to install Pypy. I downloaded the Win64 .zip and unzipped it to a convenient location. I then of course needed to run pip to install some modules. Pip was not included, but a helpful message in the interpreter, as well as [the Pypy documentation](https://doc.pypy.org/en/latest/install.html), recommended to run pypy with `-m ensurepip` to install a pip. Fine.

When I did this, I got a long and unhelpful exception, ending with "The parameter is incorrect".

image

After wasting some time fiddling, it occurred to me that WSL could be somehow causing the problem, so I tried the command in a Windows CMD.EXE shell. This worked on the first try:

```
C:\Users\Andi>work\usr\pypy3.10-v7.3.12-win64\pypy.exe -m ensurepip
Looking in links: c:\Users\Andi\AppData\Local\Temp\tmpue1h28og
Processing c:\users\andi\appdata\local\temp\tmpue1h28og\setuptools-65.5.0-py3-none-any.whl
Processing c:\users\andi\appdata\local\temp\tmpue1h28og\pip-23.0.1-py3-none-any.whl
Installing collected packages: setuptools, pip
  WARNING: The scripts pip3.10.exe and pip3.exe are installed in 'C:\Users\Andi\work\usr\pypy3.10-v7.3.12-win64\Scripts' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed pip-23.0.1 setuptools-65.5.0
```

### Expected behavior

Once an exe has been invoked inside WSL, it's just a normal Windows app running on normal Windows. So pypy should work the same if invoked from WSL Bash as if invoked from CMD.EXE, especially for a basic/"install process" kind of task like `-m ensurepip`.

If basic pypy functionality like ensurepip requires something special about the environment in which it is run— it's possible for example that my WSL was somehow creating janked environment variables, I don't know— then it should print a more helpful error message than "invalid parameter", and the environment requirements should be documented in the install instructions.

### Configuration

I am using Windows 10 Version 10.0.19045 Build 19045. My WSL is a WSL1 VM running Ubuntu 20.04.6 LTS. My pypy is pypy3.10-v7.3.12-win64. I have a file in which I have saved a copy of os.environ of pypy when invoked from WSL and a copy of same when invoked from CMD.EXE, in case this is helpful for debugging this error, but I would prefer not to post this file on a public bug report for privacy reasons.

nulano:
Fixes #4003.

Based on https://github.com/python/cpython/blob/v2.7.18/PC/getpathp.c#L414-L415, `sys.executable` should always be initialized from `GetModuleFileName` on win32 if possible.

### DIFF
--- a/extra_tests/test_startup.py
+++ b/extra_tests/test_startup.py
@@ -7,3 +7,23 @@ def test_platform_not_imported():
     modules = [x.strip(' "\'') for x in out.strip().strip('[]').split(',')]
     assert 'platform' not in modules
     assert 'threading' not in modules
+
+
+def test_executable_win32():
+    # issue #4003
+    import os
+    import subprocess
+    import sys
+
+    out = subprocess.check_output([r'*', '-c',
+         'import sys; print (repr(sys.executable))'], universal_newlines=True, executable=sys.executable)
+    if sys.platform == 'win32':
+        assert out.strip() == repr(sys.executable)
+
+        fake_executable = r'C:\Windows\System32\cmd.exe'
+        assert os.path.isfile(fake_executable)  # should exist always
+        out = subprocess.check_output([fake_executable, '-c',
+            'import sys; print (repr(sys.executable))'], universal_newlines=True, executable=sys.executable)
+        assert out.strip() == repr(sys.executable)
+    else:
+        assert out.strip() == repr('')
--- a/pypy/module/sys/initpath.py
+++ b/pypy/module/sys/initpath.py
@@ -149,6 +149,14 @@ def compute_stdlib_path_maybe(state, prefix):
 
 @unwrap_spec(executable='fsencode')
 def pypy_find_executable(space, executable):
+    if _WIN32:
+        module_filename = pypy_init_executable()
+        if module_filename:
+            module_path = rffi.charp2str(module_filename)
+            pypy_init_free(module_filename)
+            module_path = rpath.rabspath(module_path)
+            if _exists_and_is_executable(module_path):
+                return space.newtext(module_path)
     return space.newtext(find_executable(executable))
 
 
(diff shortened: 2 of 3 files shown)


extra_tests/test_startup.py
```diff
@@ -7,3 +7,23 @@ def test_platform_not_imported():
...
+    out = subprocess.check_output([r'*', '-c',
+         'import sys; print (repr(sys.executable))'], universal_newlines=True, executable=sys.executable)
+    if sys.platform == 'win32':
+        assert out.strip() == repr(sys.executable)
+
+        fake_executable = r'C:\Windows\System32\cmd.exe'
+        assert os.path.isfile(fake_executable)  # should exist always
+        out = subprocess.check_output([fake_executable, '-c',
+            'import sys; print (repr(sys.executable))'], universal_newlines=True, executable=sys.executable)
+        assert out.strip() == repr(sys.executable)
```
nulano, line 27: Checking `extra_tests/test_startup.py` with CPython, this will need to be changed for py3.11.

mattip:
Thanks @nulano

Issue #3655, opened 2022-01-29
del x; assert "x" not in locals() -> assertion fails


gitlab-importer:
In Heptapod by @njsmith on Jan 29, 2022, 19:34

This test script passes on CPython, but the `assert` fires on PyPy (tested with pypy38 7.3.7 on linux, using the official binary from pypy.org)

```python
def main():
    x = object()

    def foo():
        print(x)

    locals()
    del x
    assert "x" not in locals()

main()
```

PR #4829, opened 2024-01-04
(cfbolz, narpfel) generate shorter bytecode for `namedtuple._replace`

### LINKED ISSUE #3883: `namedtuple._replace` is slow
In Heptapod by @narpfel on Jan 2, 2023, 21:29

`namedtuple._replace` and `NamedTuple._replace` are a *lot* slower than creating a new instance with all fields manually spelled out:

```python
from collections import namedtuple
from typing import NamedTuple


class T(NamedTuple):
    a: int
    b: int
    c: int
    d: int
    e: int
    f: int


T2 = namedtuple("T2", "a, b, c, d, e, f")


def replace(t, n):
    for _ in range(n):
        t = t._replace(a=t.c + 1, c=t.a + 2)
    return t


def create_new(t, n, cls):
    for _ in range(n):
        t = cls(a=t.c + 1, b=t.b, c=t.a + 2, d=t.d, e=t.e, f=t.f)
    return t
```

```
In [1]: %timeit replace(T(0, 0, 0, 0, 0, 0), 100_000)
51.1 ms ± 1.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [2]: %timeit create_new(T(0, 0, 0, 0, 0, 0), 100_000, T)
312 µs ± 5.88 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [3]: %timeit replace(T2(0, 0, 0, 0, 0, 0), 100_000)
53.2 ms ± 408 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %timeit create_new(T2(0, 0, 0, 0, 0, 0), 100_000, T2)
2.49 ms ± 24.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```console
$ pypy3 -V
Python 3.9.16 (feeb267ead3e6771d3f2f49b83e1894839f64fb7, Dec 31 2022, 19:17:39)
[PyPy 7.3.11 with GCC 12.2.0]
```

I don’t know whether this is fixable, but I thought I’d report it anyways.

narpfel:
Using attribute access for accessing the `namedtuple`’s fields in the generated code leads to method calls to `__getitem__`, which make traces longer and require more work from the JIT.

Most of the time, the overhead of the attribute access is optimised away by the JIT, but when the `namedtuple` is large and there are many `_replace` calls, it can prevent inlining and thus further optimisations.

This comment https://github.com/pypy/pypy/issues/3883#issuecomment-1872112434 shows the difference in traced bytecode, along with an example that shows a noticeable performance difference.

By the way, the issue numbers seem to have shifted during the move to GitHub. #3883 was issue 3884 on heptapod.

Resolves #3883.

### DIFF
--- a/lib-python/3/collections/__init__.py
+++ b/lib-python/3/collections/__init__.py
@@ -457,8 +457,8 @@ def namedtuple(typename, field_names, *, rename=False, defaults=None, module=Non
     else:
         star = ""
     arguments = "\n".join(
-        f"            {field}=_self.{field} if {field} is _not_given else {field},"
-        for field in field_names
+        f"            {field}=_self[{i}] if {field} is _not_given else {field},"
+        for i, field in enumerate(field_names)
     )
     code = f"""\
 _not_given = object()


mattip:
Darn, it seems that [issue 3655 from heptapod](https://foss.heptapod.net/pypy/pypy/-/issues/3655) (#3655 does not match) did not get transfered, so everything from there up is off-by-one.

cfbolz:
thanks for persevering on this, @narpfel!

PR #4832, opened 2024-01-05
add some .gitignores and fix links in FAQ

mattip:
- ignore some translation and doc build artifacts
- fix links added in #4831


mattip:
whoohoo, the docs preview is live.

PR #4834, opened 2024-01-06
some small cleanups in rpython/: remove useless backendopts, speed up C tests, fix memory error check

cfbolz:


cfbolz:
@olliemath in addition to removing some very old code, I tweaked the rpython test running a little in this PR, to try to get the translator/c test runtimes down somewhat.

olliemath:
Nice! I think the failure of `rpython/memory/test/test_transformed_gc.py::TestIncrementalMiniMarkGC::()::test_string_concatenation` looks like flake? But perhaps points to a real issue..

cfbolz:
> Nice! I think the failure of `rpython/memory/test/test_transformed_gc.py::TestIncrementalMiniMarkGC::()::test_string_concatenation` looks like flake? But perhaps points to a real issue..

it was a real (if very weird) problem, I think. Trying again...

PR #4835, opened 2024-01-06
RPython: add support for from __future__ import print_function

olliemath:
This is mostly selfish because my code editor no longer supports the python2 print syntax. It makes both of the following valid rpython snippets (so long as they live in separate files):

```python
from __future__ import print_function

print("oh no!", end="")
```

```python
print "foo",
```

Have to admit I also have my eye on making "six-compatible" python valid rpython - for future editor, linter support etc.

### DIFF
--- a/rpython/flowspace/flowcontext.py
+++ b/rpython/flowspace/flowcontext.py
@@ -1,6 +1,6 @@
 """Implements the core parts of flow graph creation.
 """
-
+from __future__ import print_function
 import sys
 import collections
 import types
@@ -13,7 +13,7 @@ from rpython.flowspace.model import (Constant, Variable, Block, Link,
     c_last_exception, const, FSException)
 from rpython.flowspace.framestate import FrameState
 from rpython.flowspace.specialcase import (rpython_print_item,
-    rpython_print_newline)
+    rpython_print_newline, rpython_print_end)
 from rpython.flowspace.operation import op
 from rpython.flowspace.bytecode import BytecodeCorruption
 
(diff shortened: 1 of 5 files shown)


mattip:
I was trying to get documentation building working on python3 (currently it requires python2) and did some "six" changes in #4807. Maybe someone would like to continue that, taking into account that the shortest path to fixing the doc builds might be [this comment](https://github.com/pypy/pypy/pull/4807#issuecomment-1872611852)

Here is the formula for building documentation

```
python3 -m venv /tmp/venv3
source /tmp/venv3/bin/activate
cd rpython/doc   # or pypy/doc
python3 -m pip install requirements.txt
make html
```

cfbolz:
makes sense to be able to use `print_function` in rpython!

the implementation looks reasonable. it's a bit annoying that we cannot use the mechanisms in `flowspace.specialcase` to implement this, but inventing such a mechanism for a single builtin function seems overkill.

however, I'd like some more tests. the print statement in rpython is undertested, but I think we should be more thorough for the new functionality. you added a target to test this, but that's a manual step and it should really be automated. how about:
- a unit test for the rpython functions, simply calling them directly. you can monkeypatch `os.write` to check that the correct writes occur
- a unit test in `rtyper.test.test_rbuiltin` to check that the new functions are rpython and work correctly. (I don't quite know what the best way to get the actual output is. either monkeypatching `os.write` too, or you could capture the real file descriptors).

make sense?

olliemath:
> makes sense to be able to use `print_function` in rpython!
> 
> the implementation looks reasonable. it's a bit annoying that we cannot use the mechanisms in `flowspace.specialcase` to implement this, but inventing such a mechanism for a single builtin function seems overkill.
> 
> however, I'd like some more tests. the print statement in rpython is undertested, but I think we should be more thorough for the new functionality. you added a target to test this, but that's a manual step and it should really be automated. how about:
> 
>     * a unit test for the rpython functions, simply calling them directly. you can monkeypatch `os.write` to check that the correct writes occur
> 
>     * a unit test in `rtyper.test.test_rbuiltin` to check that the new functions are rpython and work correctly. (I don't quite know what the best way to get the actual output is. either monkeypatching `os.write` too, or you could capture the real file descriptors).
> 
> 
> make sense?

sure I'll give the tests a go .. there are a few cases with the print function that I might add in there too (like `print("", end="\nwhy would you do this?\r\n")`)

cfbolz:
@olliemath Oops, did you mean to delete/close this?

Issue #1220, opened 2012-07-18
Improving readability of generated .c code


gitlab-importer:
In Heptapod by bitbucket_importer on Jul 19, 2012, 24:25

*Created originally on Bitbucket by [dmalcolm (David Malcolm)](https://bitbucket.org/[7.3.7](https://github.com/gitlab-importer/pypy/milestone/7)B0a02c7d2-42a5-42bc-97b1-95a9dbf7722a%7D/)*

**From (bugs.pypy.org)**

 *Attachments:* [more-readable-c-code-2012-07-18-001.patch](https://pypy-attachments.s3.amazonaws.com/736937129/dd21f233b169a105dfd0dc67f1ee07e586a71a68159b9107ebf8e026beee5923/more-readable-c-code-2012-07-18-001.patch)[testing_1.c](https://pypy-attachments.s3.amazonaws.com/736937129/a6e0c268b64b4b1f20dfc5aa7ed6787e60d15f90f65bed3b3c79ca8baab7c1dd/testing_1.c)

PR #4837, opened 2024-01-07
RPYTHON: embed rpython code line as a comment in the generated C

mattip:
This started as a patch to #1220. I am trying to update it.

### DIFF
--- a/rpython/flowspace/flowcontext.py
+++ b/rpython/flowspace/flowcontext.py
@@ -8,9 +8,10 @@ import __builtin__
 
 from rpython.tool.error import source_lines
 from rpython.rlib import rstackovf
+from rpython.rtyper.lltypesystem import lltype
 from rpython.flowspace.argument import CallSpec
 from rpython.flowspace.model import (Constant, Variable, Block, Link,
-    c_last_exception, const, FSException)
+    c_last_exception, const, FSException, SpaceOperation)
 from rpython.flowspace.framestate import FrameState
 from rpython.flowspace.specialcase import (rpython_print_item,
     rpython_print_newline, rpython_print_end)
@@ -153,12 +154,18 @@ class Replayer(Recorder):
 
     def __init__(self, block, booloutcome, nextreplayer):
         self.crnt_block = block
-        self.listtoreplay = block.operations
+        # We don't care about comments when replaying; they
+        # are likely to be emitted at different times:
+        self.listtoreplay = [op for op in block.operations
+                             if op.opname != 'comment']
         self.booloutcome = booloutcome
         self.nextreplayer = nextreplayer
         self.index = 0
 
     def append(self, operation):
+        # Ignore comments when verifying the replay:
+        if operation.opname == 'comment':
+            return
         operation.result = self.listtoreplay[self.index].result
         assert operation == self.listtoreplay[self.index], (
             '\n'.join(["Not generating the same operation sequence:"] +
@@ -374,6 +381,28 @@ class FlowContext(object):
 
     def record(self, spaceop):
         spaceop.offset = self.last_offset
+
+        # Add a comment giving the source-code location of this operation
+        # e.g. 'is_perfect_number:35 :     return n == sum'
+        try:
+            code = self.pycode
+            linenum = self.last_offset
+            src = self.graph.source.split('\n')[linenum]
+            text = '%s:%d : %s' % (code.co_name, linenum, src)
+            comment = SpaceOperation('comment',
+                                     [Constant(text, concretetype=lltype.Void)],
+                                     Variable())
+            self.recorder.append(comment)
+        except IOError:
+            # can't read source file.  This could be due to code built
+            # on-the-fly at translation time using exec (e.g. by
+            # look_inside_iff)
+            pass
+        except Exception as e:
+            print(e)
+            import pdb;pdb.set_trace()
+            raise
+        # Add the operation itself:
         self.recorder.append(spaceop)
 
     def do_op(self, op):
--- a/rpython/flowspace/flowcontext.py
+++ b/rpython/flowspace/flowcontext.py
@@ -154,18 +154,12 @@ class Replayer(Recorder):
 
     def __init__(self, block, booloutcome, nextreplayer):
         self.crnt_block = block
-        # We don't care about comments when replaying; they
-        # are likely to be emitted at different times:
-        self.listtoreplay = [op for op in block.operations
-                             if op.opname != 'comment']
+        self.listtoreplay = block.operations
         self.booloutcome = booloutcome
         self.nextreplayer = nextreplayer
         self.index = 0
 
     def append(self, operation):
-        # Ignore comments when verifying the replay:
-        if operation.opname == 'comment':
-            return
         operation.result = self.listtoreplay[self.index].result
         assert operation == self.listtoreplay[self.index], (
             '\n'.join(["Not generating the same operation sequence:"] +
--- a/rpython/flowspace/flowcontext.py
+++ b/rpython/flowspace/flowcontext.py
@@ -381,28 +381,6 @@ class FlowContext(object):
 
     def record(self, spaceop):
         spaceop.offset = self.last_offset
-
-        # Add a comment giving the source-code location of this operation
-        # e.g. 'is_perfect_number:35 :     return n == sum'
-        try:
-            code = self.pycode
-            linenum = self.last_offset
-            src = self.graph.source.split('\n')[linenum]
-            text = '%s:%d : %s' % (code.co_name, linenum, src)
-            comment = SpaceOperation('comment',
-                                     [Constant(text, concretetype=lltype.Void)],
-                                     Variable())
-            self.recorder.append(comment)
-        except IOError:
-            # can't read source file.  This could be due to code built
-            # on-the-fly at translation time using exec (e.g. by
-            # look_inside_iff)
-            pass
-        except Exception as e:
-            print(e)
-            import pdb;pdb.set_trace()
-            raise
-        # Add the operation itself:
         self.recorder.append(spaceop)
 
     def do_op(self, op):
--- a/rpython/flowspace/flowcontext.py
+++ b/rpython/flowspace/flowcontext.py
@@ -8,7 +8,6 @@ import __builtin__
 
 from rpython.tool.error import source_lines
 from rpython.rlib import rstackovf
-from rpython.rtyper.lltypesystem import lltype
 from rpython.flowspace.argument import CallSpec
 from rpython.flowspace.model import (Constant, Variable, Block, Link,
     c_last_exception, const, FSException, SpaceOperation)
--- a/rpython/flowspace/model.py
+++ b/rpython/flowspace/model.py
@@ -275,7 +275,27 @@ class Block(object):
 
     view = show
 
-
+    def get_base_label(self, blocknum):
+        # Generate a more friendly C label for this block
+        if self.operations:
+            txt = "block"
+        elif (not self.exits) and len(self.inputargs) == 1:
+            txt = "return_block"
+        elif (not self.exits) and len(self.inputargs) == 2:
+            txt = "raise_block"
+        else:
+            txt = "block"
+        return '%s%d' % (txt, blocknum)
+    
+    def firstop(self):
+        """
+        Get the first (non-comment) operation
+        """
+        for op in self.operations:
+            if op.opname != 'comment':
+                return op
+                
+ 
 class Variable(object):
     __slots__ = ["_name", "_nr", "annotation", "concretetype"]
 
--- a/rpython/flowspace/model.py
+++ b/rpython/flowspace/model.py
@@ -170,7 +170,8 @@ class Link(object):
 
 class Block(object):
     __slots__ = """inputargs operations exitswitch
-                exits blockcolor generation""".split()
+                exits blockcolor generation
+                source_func source_line""".split()
 
     def __init__(self, inputargs):
         self.inputargs = list(inputargs)  # mixed list of variable/const XXX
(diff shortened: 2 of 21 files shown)


rpython/flowspace/flowcontext.py
```diff
@@ -374,6 +381,28 @@ def maybe_merge(self):
...
     def record(self, spaceop):
         spaceop.offset = self.last_offset
+
+        # Add a comment giving the source-code location of this operation
+        # e.g. 'is_perfect_number:35 :     return n == sum'
+        try:
+            code = self.pycode
+            linenum = self.last_offset
+            src = self.graph.source.split('\n')[linenum]
+            text = '%s:%d : %s' % (code.co_name, linenum, src)
```
mattip, line 391: This part fails. self.pycode is the compiled bytecode, and the `last_offset` relates to a bytecode offset, not a code linenum. Do we have any information about the line number in the source code anywhere?

rpython/flowspace/flowcontext.py
```diff
@@ -374,6 +381,28 @@ def maybe_merge(self):
...
+            self.recorder.append(comment)
+        except IOError:
+            # can't read source file.  This could be due to code built
+            # on-the-fly at translation time using exec (e.g. by
+            # look_inside_iff)
+            pass
+        except Exception as e:
+            print(e)
+            import pdb;pdb.set_trace()
+            raise
```
mattip, line 404: Remove this when the comment creation works
```suggestion
```

rpython/flowspace/flowcontext.py
cfbolz: yes, you can call `offset2lineno(code, self.last_offset)` to get the line number. the function is in `rpython.tool.error`.

cfbolz:
cool! pitfalls that I can think off:
- don't know what the rtyper does with the comments?
- you should make sure that the comments get a weight of 0 in inlining decisions, by adding them to `translator.backendopt.inline.OP_WEIGHTS`.

you should probably add some unit tests to:
- the flowspace to check that the right comments are generated
- the rtyper to check that it doesn't remove the comments
- the inlining tests to make sure that comments get inlined correctly, and that they don't influence inlining decisions.

mattip:
I refactored this
- added tests
- made sure inlined functions still show the comments
- left two types of source: at the top of a function, and each block that has a function line
- added a cache to minimize the slowdown in the source_c stage to ~10%.

I also ran a full translation. The result of `du -ch *.c` in the source directory grows from 451MB to 511MB.

pypy/interpreter/pycode.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, space,  argcount, nlocals, stacksize, flags,
         self.hidden_applevel = hidden_applevel
         self.magic = magic
         self._signature = make_signature(self)
+        self._cached_source = None
```
cfbolz: the changes to pypy/interpreter/pycode.py look all wrong. this is an rpython change, there shouldn't be any effects on the pypy interpreter

rpython/flowspace/flowcontext.py
```diff
@@ -153,12 +154,18 @@ class Replayer(Recorder):
 
     def __init__(self, block, booloutcome, nextreplayer):
         self.crnt_block = block
-        self.listtoreplay = block.operations
+        # We don't care about comments when replaying; they
+        # are likely to be emitted at different times:
+        self.listtoreplay = [op for op in block.operations
+                             if op.opname != 'comment']
```
cfbolz: let's add a method `noncommentops` or something to the Block class and call it here. it's bound to be useful in more places

rpython/translator/backendopt/merge_if_blocks.py
```diff
@@ -5,11 +5,13 @@
 
 
 def is_chain_block(block, first=False):
-    if len(block.operations) == 0:
+    noncommentops = [op for op in block.operations
```
cfbolz: call the new noncommentops method here

rpython/translator/c/test/test_genc.py
```diff
@@ -637,5 +680,87 @@ def main(x):
...
+    # The generated code should use a switch statement:
+    assert 'switch' in c_fn_src
+    assert 'case' in c_fn_src
+    assert 'default' in c_fn_src
+    
+    # The generated code for "a / b" inlines ll_int_py_div from rint.py,
+    # so its source comments should appear:
+    assert 'if y < 0: u = p - x' in c_fn_src
+
+def test_inlining_c_source():
```
cfbolz, line 793: there should be a unit-test for the correct handling of comments SpaceOperations in `backendopt/test/test_inline.py` too

rpython/flowspace/model.py
```diff
@@ -170,7 +170,8 @@ def show(self):
 
 class Block(object):
     __slots__ = """inputargs operations exitswitch
-                exits blockcolor generation""".split()
+                exits blockcolor generation
+                source_func source_line""".split()
```
cfbolz, line 174: I'm a bit confused in general here. there seem to be several mechanism for attaching source information to graphs. 1) attributes on Block 2) `comment` SpaceOperations 3) `source_line`/`source_func` attribute on SpaceOperations. which of these are used? for what?

later looking at _annotate_blocks_with_source I suspect it's option 1). is that true? if yes, I think we should remove the other mechanisms from this PR.

rpython/translator/simplify.py
```diff
@@ -296,8 +296,14 @@ def rename_op(op):
...
                 return op
+            target_sf = getattr(link.target, 'source_func', None)
+            target_sl = getattr(link.target, 'source_line', None)
             for op in link.target.operations:
-                link.prevblock.operations.append(rename_op(op))
+                renamed = rename_op(op)
+                if target_sf is not None and not hasattr(renamed, 'source_func'):
+                    renamed.source_func = target_sf
+                    renamed.source_line = target_sl
+                link.prevblock.operations.append(renamed)
```
cfbolz, line 306: this change to `simplify.py` needs a unit test in `test/test_simplify.py`

rpython/translator/simplify.py
```diff
@@ -1054,6 +1060,36 @@ def run(self, vlist, vmeth, appendblock):
                     newblock.operations.append(hint)
         # done!
 
+def merge_comments(graph):
```
cfbolz: this function needs unit tests. or maybe it can just be removed, if we use attributes on blocks instead?

rpython/translator/translator.py
```diff
@@ -18,6 +18,29 @@
 
 log = AnsiLogger("flowgraph")
 
+def _annotate_blocks_with_source(graph):
```
cfbolz, line 21: this logic needs unit tests in `test_translator.py`

rpython/translator/c/funcgen.py
```diff
@@ -180,6 +192,33 @@ def cfunction_body(self):
             if extra_enter_text:
                 yield extra_enter_text
         graph = self.graph
+        # Try to print python source code:
+        if hasattr(graph, 'func'):
```
cfbolz, line 256: the logic here has some overlap with the code in rpython/tool/error.py and rpython/tool/sourcetools.py. particularly the logic in sourcetools.py to stash extra information into the code object for dynamically generated RPython source code should be integrated (because we do that quite a bit).

rpython/translator/backendopt/inline.py
```diff
@@ -257,6 +257,11 @@ def copy_block(self, block):
                 self.passon_vars(block))
         newblock = Block(args)
         self._copied_blocks[block] = newblock
+        if hasattr(block, 'source_func'):
```
cfbolz: this logic needs a unit test.

also, there is similar block creation logic in unsimplify.py, which might need similar changes (maybe in a helper method on Block)?

rpython/translator/c/funcgen.py
```diff
@@ -28,6 +29,17 @@ def make_funcgen(graph, db, exception_policy, functionname):
...
 
+# Cache (func -> (src_lines, startline)) so getsourcelines() is called at most
+# once per function rather than once per block/operation.
+_source_lines_cache = {}
+
+def escape_c_comments(py_src):
+    # Escape C comments within RPython source, to avoid generating bogus
+    # comments in our generated C source:
+    py_src = py_src.replace('/*', '')
+    py_src = py_src.replace('*/', '')
```
XmiliaH: This seems problematic when dealing with constructs such as `**//` which will result in `*/` after the replacement.

rpython/flowspace/model.py
mattip: Well spotted. 2 was the earlier attempt and is just dead code. I will remove the dead code.  What remains:
1. source_func/source_line on **Block** objects
   - set by `_annotate_blocks_with_source`, 
   - copied on inlining/splitting, 
   - read by `funcgen.py:gen_block` to emit a block-level /* fname:N: line */
  comment
3. source_func/source_line on **SpaceOperation** objects
   - propagated by `simplify.py:join_blocks` when blocks are merged, 
   - read by `funcgen.py:gen_block` to emit per-operation source comments
  after block boundaries have been erased

rpython/translator/c/test/test_genc.py
mattip: Added more tests

rpython/translator/c/funcgen.py
mattip: Refactored and more tests for generated code added

rpython/translator/simplify.py
mattip: Tests added

rpython/translator/translator.py
mattip: Tests added

rpython/translator/c/funcgen.py
```diff
@@ -58,6 +59,16 @@ def make_funcgen(graph, db, exception_policy, functionname):
...
     return FunctionCodeGenerator(graph, db, exception_policy, functionname)
 
+# Cache (func -> (src_lines, startline)) so getsourcelines() is called at most
+# once per function rather than once per block/operation.
+_source_lines_cache = {}
+
+def escape_c_comments(py_src):
+    # Escape C comment delimiters within RPython source.
+    py_src = py_src.replace('/*', '/ *')
+    py_src = py_src.replace('*/', '* /')
```
mattip, line 69: The comment about these lines got lost in the refactoring. Note how `/*` becomes `/ *` to not break C code.

mattip (commented):
I think I related to all the review comments.

rpython/translator/c/funcgen.py
XmiliaH: This looks better to me.
I also thought about the end sequence with an escaped newline
```
*\
/
```
but this should not give a problem here as all lines are handled separately and get at least a ` *` prefix which will escape that sequence.

rpython/flowspace/flowcontext.py
```diff
@@ -8,9 +8,10 @@
 
 from rpython.tool.error import source_lines
 from rpython.rlib import rstackovf
+from rpython.rtyper.lltypesystem import lltype
```
cfbolz: this change is churn, the extra imports in flowcontext.py are not needed.

rpython/rtyper/lltypesystem/lloperation.py
```diff
@@ -587,18 +587,18 @@ def specialize_call(self, hop):
...
+    'debug_start':             LLOp(canrun=True),
+    'debug_stop':              LLOp(canrun=True),
+    'have_debug_prints':       LLOp(canrun=True),
+    'have_debug_prints_for':   LLOp(canrun=True),
+    'debug_offset':            LLOp(canrun=True),
+    'debug_flush':             LLOp(canrun=True),
+    'debug_assert':            LLOp(tryfold=True),
+    'debug_assert_not_none':   LLOp(tryfold=True),
+    'debug_fatalerror':        LLOp(canrun=True),
+    'debug_llinterpcall':      LLOp(canraise=(Exception,)),
```
cfbolz: this change is churn, just whitespace changes.

rpython/tool/test/test_sourcetools.py
```diff
@@ -55,3 +56,63 @@ def {name}({arglist}):
...
+def test_getsourcelines_dynamic_function_returns_none():
+    # Functions compiled from a string literal (no real source file) return None
+    # rather than raising.
+    import types
+    code = compile('def f(x): return x', '<string>', 'exec')
+    globs = {}
+    exec code in globs
+    f = globs['f']
+    assert getsourcelines(f) is None
+
```
cfbolz, line 93: let's add this (passing) test (it needs an import of compile2 at the top):

```python
def test_getsourcelines_dynamic_function_compile2_works():
    # Functions compiled from a string literal compiled with compile2 can
    # produce a source line (this is the point of compile2, and why we use it
    # eg in the objspace).
    import types
    code = compile2('def f(x): return x', '<string>', 'exec')
    globs = {}
    exec code in globs
    f = globs['f']
    assert getsourcelines(f) == (['def f(x): return x\n'], 1)
```

cfbolz (commented):
I'm pretty happy with this now. three minor comments below.

rpython/flowspace/flowcontext.py
mattip: removed

rpython/rtyper/lltypesystem/lloperation.py
mattip: reverted

rpython/tool/test/test_sourcetools.py
mattip: done

mattip (commented):
I made the requested changes

cfbolz:
Merged. Thanks @mattip for pushing this through!

Issue #4826, opened 2024-01-04
AssertionError and abort on setting `__bases__` in PyPy 3.9 v7.3.14


YannickJadoul:
The newly released version of PyPy 3.9 crashes during the import of the following pybind11 extension module:

```c++
#include <pybind11/pybind11.h>

class A {};
class B {};

namespace py = pybind11;

PYBIND11_MODULE(mwe_pypy, m) {
    py::class_<A> classA(m, "A");
    py::class_<B> classB(m, "B");
    classB.attr("__bases__") = py::make_tuple(classA);
}
```
with the following message:
```
RPython traceback:
  File "pypy_interpreter_2.c", line 58110, in GetSetProperty_descr_property_set
  File "pypy_module_cpyext_7.c", line 35058, in fill_slot__c_tp_as_buffer_c_bf_getbuffer
Fatal RPython error: AssertionError
```

Strangely enough, this only happens on PyPy 3.9 v7.3.14. Both PyPy 3.10 v7.3.14 and PyPy 3.9 v7.3.13 work perfectly fine.

---

I haven't managed to further reduce this example by getting rid of pybind11, but hoped reporting this might already help.
Here is the current minimal example from above, with CMake and pybind11 included: [mwe_pypy.zip](https://github.com/pypy/pypy/files/13824744/mwe_pypy.zip)

It can be built and run with:
```sh
cmake -S . -B build -DPython_EXECUTABLE=pypy3.9-v7.3.14-linux64/bin/pypy -DCMAKE_BUILD_TYPE=Debug
cmake --build build
PYTHONPATH=build pypy3.9-v7.3.14-linux64/bin/pypy -c "import mwe_pypy"
```

<details>
<summary>And this is the gdb stack trace of the crash</summary>

```
#0  0x00007ffff45589fc in pthread_kill () from /usr/lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff4504476 in raise () from /usr/lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff44ea7f3 in abort () from /usr/lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff6a7e7f1 in pypy_debug_catch_fatal_exception () from /home/yannick/pypy3.9-v7.3.14-linux64/bin/libpypy3.9-c.so
[...]

PR #4847, opened 2024-01-10
Do not assign c-level slots when the python type has no function to assign

mattip:
Toward fixing #4826. 

This form of the test passes, there must be more to the failure.


mattip:
Fixed the test by backporting a required change from py3.10. Fixes #4826

Issue #4836, opened 2024-01-07
recent rpython work broke existing tests


mattip:
I see a nightly build has [new failures](https://buildbot.pypy.org/summary?branch=main): something about unicode no longer raising. @olliemath could you take a look?

mattip:
Strangely, the failure does not show up [on windows](https://buildbot.pypy.org/summary?builder=own-win-x86-64&branch=main)

olliemath:
Ah, about to catch a flight so might not be able to look for a bit

Yes I fixed a ValueErrror raised in one of the rpython tests (the test
asserted that casting -1 to unicode and back to long gave either
sys.maxunicode or -1) but it was failing because a value error was raised
when run under pypy

This test looks like it expects the opposite semantics: a ValueError to be
raised for out of range integers. I suspect only one or other of the tests
can pass and the other should be skipped or xfailed

I guess the previous behaviour was the expected behaviour .. so we should
just revert my "fix" and skip the failing rpython test

mattip:
Not critical, it can wait.

I am not sure we should be expecting the ValueError: I don't see any other effects of this change. If raising was desired behaviour, I would expect to see more failures. I merged to py3.9 let's see if anything new fails

olliemath:
Ah true - cpython2 doesn't raise a ValueError with this snippet - I assume compatibility with what cpython does is the desired behaviour:

```python
>>> from array import array
>>> input_unicode = u'\x01\u263a\x00\ufeff'
>>> a = array('u', input_unicode)
>>> b = array('u', input_unicode)
>>> b.byteswap()
>>> assert b[2] == u'\u0000'
>>> assert a != b
>>> b.itemsize
4
>>> b[0]
u'\U01000000'
```

So I think we should fix that test to check the cpython behaviour

mattip:
When merging thi with py3.9 there is a [new untranslated failure](https://buildbot.pypy.org/summary/longrepr?testname=AppTestUnicodeObject.%28%29.test_invalid&builder=own-linux-x86-64&build=10020&mod=pypy.module.cpyext.test.test_unicodeobject). The same test passes after translation. I am fine if we adjust the tests to all pass and call it a corner-case incompatibility.

olliemath:
So playing around with python3 it _does_ raise a ValueError here (namely `ValueError: character U+1000000 is not in range [U+0000; U+10ffff]` ) - my current plan is to revert the rpython commit and change the test in a slightly different way

[...]

PR #4851, opened 2024-01-13
 revert 6906755 and skip untranslated tests 

mattip:
fixes some untranslated tests

### DIFF
--- a/pypy/module/_rawffi/test/test__rawffi.py
+++ b/pypy/module/_rawffi/test/test__rawffi.py
@@ -6,7 +6,7 @@ from pypy.module._rawffi.tracker import Tracker
 
 import os, sys, py
 
-class AppTestFfi:
+class AppTestFfi(object):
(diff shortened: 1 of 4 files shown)


mattip:
cc @olliemath

mattip:
Closes #4836

olliemath:
Nice, thanks for this - I've been thinking on it this week. My conclusion was that it would actually be more consistent to raise a ValueError for all instances of `cast(lltype.UniChar, -1)` (regardless of platform).

This might break some very niche use-cases, but already this functionality isn't being used during pypy translation. And the bonus is that users don't have to worry about `sys.maxunicode` or `sys.executable` to know whether this operation is valid.

Happy to defer that to a future pull request.

mattip:
It seems like such a niche use case that whatever we choose to do should be OK. There is the small issue that PyPy2 cannot index this array
```
array.array('u' , array.array("I", (0xffffffff,)).tostring())
```

where CPython2 can

PR #4854, opened 2024-01-14
RPython: add support for from __future__ import print_function

olliemath:
Recreates https://github.com/pypy/pypy/pull/4835 (after I deleted + recreated my fork)

This is mostly selfish because my code editor no longer supports the python2 print syntax. It makes both of the following valid rpython snippets (so long as they live in separate files):

```python
from __future__ import print_function

print("oh no!", end="")
```

```python
print "foo",
```

Have to admit I also have my eye on making "six-compatible" python valid rpython - for future editor, linter support etc.

### DIFF
--- a/rpython/annotator/test/test_annrpython.py
+++ b/rpython/annotator/test/test_annrpython.py
@@ -1,4 +1,4 @@
-from __future__ import with_statement
+from __future__ import with_statement, print_function
 import py.test
 import sys
 from collections import OrderedDict
@@ -997,7 +997,7 @@ class TestAnnotateTestCase:
         # check that the list produced by range() is not mutated or resized
         graph = graphof(a, snippet.harmonic)
         all_vars = set().union(*[block.getvariables() for block in graph.iterblocks()])
-        print all_vars
+        print(all_vars)
         for var in all_vars:
             s_value = var.annotation
             if isinstance(s_value, annmodel.SomeList):
@@ -1077,8 +1077,8 @@ class TestAnnotateTestCase:
         assert s == annmodel.SomeInteger(nonneg = True, unsigned = True)
 
 
-    def test_prebuilt_long_that_is_not_too_long(self):
-        small_constant = 12L
+    def test_prebuilt_int_that_is_not_too_long(self):
+        small_constant = 12
         def f():
             return small_constant
         a = self.RPythonAnnotator()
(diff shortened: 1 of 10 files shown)


olliemath:
> makes sense to be able to use print_function in rpython!
> 
> the implementation looks reasonable. it's a bit annoying that we cannot use the mechanisms in flowspace.specialcase to > implement this, but inventing such a mechanism for a single builtin function seems overkill.
> 
> however, I'd like some more tests. the print statement in rpython is undertested, but I think we should be more thorough for the new functionality. you added a target to test this, but that's a manual step and it should really be automated. how about:
> 
>     a unit test for the rpython functions, simply calling them directly. you can monkeypatch os.write to check that the correct writes occur
>     a unit test in rtyper.test.test_rbuiltin to check that the new functions are rpython and work correctly. (I don't quite know what the best way to get the actual output is. either monkeypatching os.write too, or you could capture the real file descriptors).
> 
> make sense?

^ from https://github.com/pypy/pypy/pull/4835

olliemath:
So I ended up adding tests for the logic of the rpython print functions (this lead me to simplify the functions and account for some edge-cases by keeping track of a little more state) plus some tests to `translate/c/tests/...` which is where the original print statement had its tests added

olliemath:
The failed memory test here is interesting - it looks unrelated - a similar failure was seen over here: https://github.com/pypy/pypy/actions/runs/7466932227/job/20319376709#step:5:627 (RuntimeError: pointer comparison with a freed structure)

mattip:
I ported the wiki to github, and discovered [this page](https://github.com/pypy/pypy/wiki/RPython-3-considerations) in which you laid out a path for running RPython with Python3 :)

olliemath:
> I ported the wiki to github, and discovered [this page](https://github.com/pypy/pypy/wiki/RPython-3-considerations) in which you laid out a path for running RPython with Python3 :)

haha yes - it's quite a long list - I believe your docs branch would tick off a few of the "simple" items

mattip:
Related to #4807. I guess the next step is to go through files with `print ...`  and convert them to use `from __future__ import print_function` and use `print(...)`

olliemath:
Yes - I'll try and keep it separate to that branch though as it will be a pain to review (100s of files with no meaningful semantic change)

mattip:
> I'll try and keep it separate to that branch though

The rpython3 branch is so out of date, perhaps it makes sense to make a new one. Can lib2to3  refactor only print statements?

PR #4865, opened 2024-01-15
[WIP] Run lib-python tests during translation

Julian:
This uses Tim's Graal script to run the "known passing" lib-python tests.

Marking this as a WIP as even though this runs (see e.g. https://github.com/Julian/pypy/actions/runs/7531739205/job/20500943171) I haven't yet stared at that to see whether it's doing the *right* thing and not just always passing now even if tests fail. I've been only taking brief looks at making progress but hopefully putting it here will make me go faster and/or let me know if any of you fine folks have preferences on how this should work beyond what I did which is just "add it to the workflow and pull in the tags".


mattip:
I have a few concerns:
- What is the compatibility of the UPL license with other licenses? Can we freely ship this without limiting our downstream users?
- The tests must be individually enabled/disabled. This seems like it would add a maintenance burden.
- The translate CI run is now taking 93 minutes. That is too long for a CI job (continuous integration), and would be more appropriate for the buildbot nightly runs. But we already run stdlib tests on the buildbots...

Julian:
Thanks! All good points of course -- some quick responses though I'm sure Carl will have opinions:

> What is the compatibility of the UPL license with other licenses? Can we freely ship this without limiting our downstream users?

From reading the terms it looks to me like yes, it's basically permitting use and modification even for commercial use. One could of course also delete the file -- whether anyone *still* has any downstream concerns though I of course don't know but yeah there's of course another path which is essentially reimplementing skipping test cases from scratch, that I guess shouldn't be prohibitively difficult to do, the annoying part is just how long it takes to run the suite as you noticed.

> The tests must be individually enabled/disabled. This seems like it would add a maintenance burden.

So the script has a way to re-run itself and regenerate the passing tests -- here I implemented using that as clicking a checkbox when running the action in workflow dispatch:

Screenshot 2024-01-16 at 9 59 43 AM

which means you then can download (and commit) updated "tags" with the passing tests.

I guess another improvement would be an "expected fail" mode where it runs all tests and *errors* if any pass when they were expected to fail, but then you need something specifically for flaky tests.

> that is too long for a CI job (continuous integration), and would be more appropriate for the buildbot nightly runs. But we already run stdlib tests on the buildbots...

(I agree of course that that's very long) -- just to list other options without a real personal preference -- we could also disable these by default and *only* run them if you enable them in the workflow dispatch I screenshotted above. We could also run them on `cron` here in GitHub actions -- the reasoning being I think it's still "useful" to take advantage of GH CI resources/infra as well as buildbot.

(Ideas or more thoughts on all the above of course welcome, as I say I've been slow at looking at this but happy to help if this is important)

mattip:
OK, I will wait for @cfbolz to comment. In the meantime:

> So the script has a way to re-run itself and regenerate the passing tests 

Thanks. Documentation needed.

> I think it's still "useful" to take advantage of GH CI resources/infra as well as buildbot.

In that case this should run windows, macos, and linux at a minimum, and be enabled for PRs against the py3.9 and py3.10 branches.

Issue #4866, opened 2024-01-16
Flaky rpython memory tests


olliemath:
These two failures:

https://github.com/pypy/pypy/actions/runs/7533803201/job/20507000160
https://github.com/pypy/pypy/actions/runs/7533317545/job/20505615167

and this success:

https://github.com/pypy/pypy/actions/runs/7538948557/job/20520450692

are with the same code (I removed a line of whitespace to retrigger the test). A similar suspected flake is here: https://github.com/pypy/pypy/actions/runs/7466932227/job/20319376709

The failures were:

```python
____________________ TestIncrementalMiniMarkGC.test_weakref ____________________

self = <rpython.memory.test.test_transformed_gc.TestIncrementalMiniMarkGC object at 0x0000000064eaf2b8>

    def test_weakref(self):
        run = self.runner("weakref")
>       res = run([])

rpython/memory/test/test_transformed_gc.py:523: 
<... snip .../>
self = <C object Struct timeval { c_tv_sec, c_tv_usec } at 0x6805d60>
other = <struct header { tid=<CombinedSymbolic GroupMemberOffset(* <rpython.rtyper.lltypesystem.llgroup.group object at 0x00000000070fed08>, 3)|279172874240> }>

    def __eq__(self, other):
        if isinstance(other, _llgcopaque):
            addressof_other = other.intval
        else:
            if not isinstance(other, lltype._parentable):
                return False
            if self._storage is None or other._storage is None:
>               raise RuntimeError("pointer comparison with a freed structure")
E               RuntimeError: pointer comparison with a freed structure
```

and 

```python
_______________ TestIncrementalMiniMarkGC.test_nursery_hash_base _______________

self = <rpython.memory.test.test_transformed_gc.TestIncrementalMiniMarkGC object at 0x0000000063bd8c98>

    def test_nursery_hash_base(self):
        res = self.runner('nursery_hash_base')
>       assert res([]) >= 195

rpython/memory/test/test_transformed_gc.py:1373: 
<... snip .../>
self = <C object Struct timeval { c_tv_sec, c_tv_usec } at 0xc888c80>
other = <struct header { tid=-42 }>

    def __eq__(self, other):
        if isinstance(other, _llgcopaque):
            addressof_other = other.intval
        else:
            if not isinstance(other, lltype._parentable):
                return False
            if self._storage is None or other._storage is None:
>               raise RuntimeError("pointer comparison with a freed structure")
E               RuntimeError: pointer comparison with a freed structure
```

[...]

PR #4867, opened 2024-01-16
use print_function in rpython

olliemath:
Related to https://github.com/pypy/pypy/pull/4807 

I avoided changing any test files and snippets in translator/goal, since we want to support old-style python2 print functions for targets, so that needs to continue being tested.

Aside from that, this is an application of the print fix from lib2to3 plus the addition of `from __future___ import print_function` in relevant files

### DIFF
--- a/rpython/__main__.py
+++ b/rpython/__main__.py
@@ -4,12 +4,13 @@ rpython <translation options> target <targetoptions>
 
 run with --help for more information
 """
+from __future__ import print_function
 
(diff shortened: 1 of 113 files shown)


olliemath:
memory test failure here is just flake mentioned in https://github.com/pypy/pypy/issues/4866

mattip:
Let's give it a try, thanks @olliemath

PR #4868, opened 2024-01-16
make test_transformed_gc tests more stable

olliemath:
this is based on a hunch that the flake in these tests is due to deferred gc of spurious entries in the header2obj WeakKeyDictionary (which would be why cpython runs of the tests are more stable) ... to test the theory I need to trigger a lot of CI runs

closes #4866

PR #4870, opened 2024-01-17
Revert "make test_transformed_gc tests more stable"

olliemath:
Reverts pypy/pypy#4868

This didn't seem to work - so something else must be at play here

### DIFF
--- a/rpython/memory/test/test_transformed_gc.py
+++ b/rpython/memory/test/test_transformed_gc.py
@@ -52,12 +52,6 @@ class GCTest(object):
     taggedpointers = False
     gchooks = None
 
-    def setup(self):
-        # This snippet is an attempt to clean the WeakKeyDict
-        # rpython.memory.gcheader.header2obj between tests
-        import gc
-        gc.collect()
-
     def setup_class(cls):
         cls.marker = lltype.malloc(rffi.CArray(lltype.Signed), 1,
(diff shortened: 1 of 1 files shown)


mattip:
Let's let this change run by the rpython nightly buildbot build, perhaps it will give us more data.

mattip:
That specific test `test_transformed_gc` does not show up at all in the buildbot runs, neither before or after #4868. Some of the buildbots (macos-arm64 macos-x86_64, windows) use pypy2 as a test runner host python, and last night's run cleared up some other failures on macos-arm64 `test.test_boehm`, so maybe you are on to something? Are there other places that also need to call the GC to clear caches?

mattip:
Here is the [link to the rpython builbot runs](https://buildbot.pypy.org/summary?builder=rpython-linux-x86-32&builder=rpython-linux-x86-64&builder=rpython-win-x86-64&builder=rpython-win-x86-32&builder=rpython-linux-aarch64&builder=rpython-macos-arm64&builder=rpython-macos-x86-64), it also appears on the top navbar of the buildbot pages

mattip:
What do you think: should we merge this or leave #4868 in?

mattip:
Closing. It seems #4868 __did__ fix the problem, the test is now reliably passing on all buildbots and in the CI action.

Issue #4860, opened 2024-01-15
Unicode dir is seen as not a dir on windows only via os.scandir...


Carreau:
This is quite recent failure on IPython, 

I'm using https://github.com/ipython/ipython/pull/14293 to debug, and already pinged `@mattip`.

It looks like somehow between 7.3.13, and 7.3.14, folder with unicode names are reported as not-a-directory on windows.

```
    def test_wierd_pypy_73713_7314():
        TMP = Path(tempfile.mkdtemp())
        td = Path(tempfile.mkdtemp(dir=TMP))
        p = Path(td / "profile_ünicode2") 
        p.mkdir(parents=True) # p should really be a dir
        assert p.is_dir() # Yep it's s dir.
        l = list(os.scandir(td)) # let's get it via scandir.
>       assert l[0].is_dir(), ("l[0]", l[0], "should be dir is not") 
E       AssertionError: ('l[0]', <DirEntry 'profile_ünicode2'>, 'should be dir is not')
E       assert False
E        +  where False = <bound method DirEntry.is_dir of <DirEntry 'profile_ünicode2'>>()
E        +    where <bound method DirEntry.is_dir of <DirEntry 'profile_ünicode2'>> = <DirEntry 'profile_ünicode2'>.is_dir
```

mattip:
@cpburnz any thoughts about what is wrong? Maybe something in the PR to handle symlinks in pypy + windows + os.scandir is a bit off?

mattip:
Debugging this, the difference comes from [W_DirEntry.check_mode](https://github.com/pypy/pypy/blob/ae536562f5b45ccbd8a0f61ad8622fb7a657d003/pypy/module/posix/interp_scandir.py#L317), which used to simply [return `self.flags`](https://github.com/pypy/pypy/blob/bd1a0d059f5e947f45cffb2c0293049ea2d30557/pypy/module/posix/interp_scandir.py#L325). In the PR to support symlinks, it was changed to call `get_stat_or_lstat`, which chokes on unicode names.

mattip:
[...]

PR #4873, opened 2024-01-19
add a test of scan_dir with symlink and unicode

mattip:
xref #4860 

@cpburnz is 508538bef the test you were thinking about?


mattip:
No, this is targeting py3.9, it should be main

cpburnz:
@mattip Yes, that's the test I was thinking of, and that should cover it.

mattip:
Thanks for the review, merging

PR #4886, opened 2024-02-04
Remove top-level __init__.py when packaging pypy

### LINKED ISSUE #4885: top-level __init__.py causes import error
As reported in https://github.com/conda-forge/pycalphad-feedstock/pull/55, somehow pytest is catching the wrong error being raised here https://github.com/pypy/pypy/blob/1fca5847f1902f76523d805ed291763b23733ccb/lib_pypy/__init__.py#L3-L4

The easiest fix is to delete the file when packaging pypy

mattip:
Fixes #4885


mattip:
Nope, targeting the wrong branch

PR #4891, opened 2024-02-08
Merge PR #57 from cffi.

arigo:
Should be tested on Windows with the automated test runners.


mattip:
Should the non-bakcend changes and tests from python-cffi/cffi#57 also be ported?

arigo:
Refresh my memory, we're not updating cffi automatically any more, correct?  Then yes, I can port the non-backend changes and the tests that are applicable.  Should I use the version 1.17?  It's not like cffi 1.17 is released yet and we don't really know if there will be more changes to do in PyPy for 1.17...

mattip:
I did merge the previous changes manually in #4849 but I think that was simpler. I don't know if the automatic tool still works, does it still think cffi's VCS is mercurial?

mattip:
> Should I use the version 1.17?

I think so. We don't really have a timeline for another PyPy release either.

arigo:
I have maaaybe re-enabled the merge of cffi, but I worry this will introduce more issues than it solves.  Even running the cffi tests is becoming messy, as there is some Python 3 syntax introduced in them.  Either the current approach works, or else I'll scrap that and try to add a few of the new tests explicitly and not worry about keeping stuff up-to-date.

arigo:
OK, I've run manually `extra_tests/cffi_tests/` on Linux on a just-translated pypy -O2.  All tests now pass, so I guess we can just merge this and look at the Windows buildbot for more.

mattip:
Thanks @arigo

PR #4894, opened 2024-02-13
extra_tests: Fix test_sqlite3 test_connection_del

loganchien:
This commit fixes the failing `test_connection_del` test case in `test_sqlite3.py`.

The test case relies on `gc.collect()` to invoke the `Connection.__del__` method.  But, the `con` list in the `open_many` function is still reachable after the first `open_many(False)` invocation.  (Note: It seems that the exception holds a backtrace, the backtrace holds the frame object, the frame object keeps the `con` list and the `Connection` objects reachable.)

This commit fixes the problem by wrapping the first invocation in a function so that the propagating exception object can be freed before we call `gc.collect()`.

### DIFF
--- a/extra_tests/test_sqlite3.py
+++ b/extra_tests/test_sqlite3.py
@@ -144,8 +144,10 @@ def test_connection_del(tmpdir):
                     con[i] = None
                     gc.collect(); gc.collect()
 
-        with pytest.raises(_sqlite3.OperationalError):
-            open_many(False)
+        def assert_fds_limit_triggers_operational_error():
+            with pytest.raises(_sqlite3.OperationalError):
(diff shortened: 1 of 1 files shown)


mattip:
Thanks, this seems right. Where did you see the test fail?

loganchien:
> Thanks, this seems right. Where did you see the test fail?

I found this when I was debugging [RISC-V JIT backend failures](https://mail.python.org/archives/list/pypy-dev@python.org/message/5JEP6WFOTFRSFR47JNHEOZRXDJ3RXCML/).  And then, it turned out x86 backend was failing when I ran it locally.  I was not quite sure why it was green on buildbot.

PR #4911, opened 2024-02-21
(untested) Win32: CFFI: pass the flags from dlopen() to LoadLibraryEx()

arigo:
https://github.com/python-cffi/cffi/pull/65


mattip:
Also see python/cpython#12302, which is the PR that changed the `ctypes.DLL` function on windows to accept a `winmode` argument. This was partially implemented in the ctypes backend for PyPy.

PR #4912, opened 2024-02-21
Partially Implemented ExceptionGroups (from PEP654)

nirit100:
Only implemented the `BaseExceptionGroup` and `ExceptionGroup` classes. No integration rn. 

Based on this implementation: [ExceptionGroup backport](https://github.com/agronholm/exceptiongroup/)

Also, there are still one FIXME and two TODOs: 
- [x] FIXME Test `test_notes_is_list_of_strings_if_it_exists` is marked as `xfail` right now because `add_note` is not available yet
- [x] TODO Test copying the `__notes__` over in the `derive` function
- [x] TODO Test whether or not copying `__cause__`, `__context__` and `__traceback__` is fine to be done in `derive`, because unlike `__notes__` that was done on the caller side in the referenced implementation.

### DIFF
--- a/extra_tests/test_exceptiongroups.py
+++ b/extra_tests/test_exceptiongroups.py
@@ -103,9 +103,7 @@ def test_fields_are_readonly():
     with pytest.raises(AttributeError):
         eg.exceptions = [OSError("xyz")]
 
-@pytest.mark.xfail(reason="TODO add_note is not available!")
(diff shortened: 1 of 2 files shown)


cfbolz:
here's the source code of derive: https://github.com/python/cpython/blob/3.11/Objects/exceptions.c#L880

and of subset: https://github.com/python/cpython/blob/3.11/Objects/exceptions.c#L920

Issue #4925, opened 2024-03-15
nanobind tests lead to gc invariant violation


cfbolz:
I've been investigating a bug reported by @wjakob [here](https://github.com/orgs/pypy/discussions/4923). it happens as random crashes on github ci. to reproduce it:

```
git clone --recursive git@github.com:wjakob/nanobind.git
cd nanobind
pypy3.10 -m venv venvpypy310
. venvpypy310/bin/activate
pip install pytest pytest-github-actions-annotate-failures typing_extensions
cmake -S . -B build -DNB_TEST_STABLE_ABI=ON -DNB_TEST_SHARED_BUILD="$(python3 -c 'import sys; print(int(sys.version_info.minor>=11))')"
cmake --build build
cd build
PYPY_GC_DEBUG=2 PYPYLOG=gc:- python --jit off -m pytest
```

wjakob:
Excellent! Is this with some nanobind functions still on the backtrace, or after an operation (e.g. type creation)?

cfbolz:
no, so far it actually seems mostly unrelated to nanobind? if I run it with `python -v` to print the imports, the only extension module that gets build before the crash is `_lzma`, which uses cffi.

cfbolz:
indeed, with the pypy3.10-lldebug that I just freshly built, I can even get this to crash?!:

```
pypy3.10 -m venv venvpypy310
. venvpypy310/bin/activate
pip install pytest
PYPY_GC_DEBUG=2 PYPY_GC_NURSERY_DEBUG=1 PYPYLOG=gc:- python --jit off -c "import pytest"
```

cfbolz:
it seems I managed to reproduce the crash in an (untranslated) unit-test of the GC. both in the form where GC debugging is turned on (then I get the assertion error) and in the form where the GC frees a still reachable object. Very annoying bug, but I'm happy to have managed to write a test.

cfbolz:
I spent a lot more time in rr today and may have found the source of the crash (as opposed to the source of the failing invariant assert, which seems to be a different issue).

Here's a [diff](https://github.com/pypy/pypy/compare/gc-arraycopy-incrementality-bug?expand=1) for my new attempt at reproducing the problem and fixing it, @arigo, could you maybe take a look?

arigo:
[...]

Issue #3959, opened 2023-07-07
Pypy segfaults during pytest test rewriting


gitlab-importer:
In Heptapod by @LilyFoote on Jul 7, 2023, 11:08

For a while now I've been seeing pypy segfault in CI fairly frequently on a work project. A few days ago I reported this to the pytest issue tracker: https://github.com/pytest-dev/pytest/issues/11168

Since then, I have further investigated the issue and have reached this reproduction: https://github.com/LilyFoote/pypy-segfault/tree/61ca92436b75bbd4dd5690b31aa5c780512fa04d which triggered the segfault here: https://github.com/LilyFoote/pypy-segfault/actions/runs/5484866146/jobs/9993059068

I have not been able to reproduce the segfault locally or on github actions own `ubuntu-latest` runner, so I think the hardware used for the `buildjet-2vcpu-ubuntu-2204` runner may be relevant. Unfortunately the information about the hardware is pretty sparse: https://buildjet.com/for-github-actions/docs/runners/hardware

The test that causes the segfault is usually `test_get_sync`, but I have seen one run where the first three (`test_get_sync`, `test_get_streaming_response` and `test_get_binary_response`) all failed.

mattip:
I think this is solved by the gc arraycopy fixes in #4928. Please reopen or open a new issue if the problem returns.

PR #4928, opened 2024-03-26
Gc arraycopy incrementality bug

cfbolz:
should address #4925 and #3959


mattip:
Cool. I guess we should think about a release

Issue #4927, opened 2024-03-20
Update for latest CPython release


mattip:
CPython did a [security release](https://discuss.python.org/t/python-3-10-14-3-9-19-and-3-8-19-is-now-available)
We should
- [ ] Update libexpat to 2.6.0. This requires updating the build workers (linux Dockerfiles, windows dependency, macos via brew). Perhaps we should do like CPython and bundle libexpat into our sources for `pypy/modules/pyexpat` python/cpython#115474 which would statically link libexpat and reduce the number of bundled shared objects from 8 to 7
- [ ] Update the stdlib to get the fixes for zipfile, tempfile, urllib, socket test, hidden pth, iso2022_jp_3 test, libexpat test, etc
- [ ] Check that we properly handle the underlying socket function changed in https://github.com/python/cpython/pull/112600
- [ ] Check that we pass the iso2022_jp_3 test https://github.com/python/cpython/pull/111780
- [ ] Use X509_STORE_get1_objects instead of X509_STORE_get0_objects in _ssl https://github.com/python/cpython/pull/115548, raising better error in _ssl https://github.com/python/cpython/issues/107077
- [ ] Update bundled OpenSSL to 3.0.11
- [ ] Update zlib 1.3.1
- [ ] Use new libexpat API https://github.com/python/cpython/pull/116272

cc @mgorny

cfbolz:
> Check that we pass the iso2022_jp_3 test https://github.com/python/cpython/pull/111780

we have vendored a (modified?) copy of `_codecs_iso2022.c`, so I think we need to do some fixes here.

mattip:
I updated py3.9 to 3.9.19

mattip:
Newly failing tests, to be investigated:
On py3.9
- [ ] lib-python.3.test.test_codecencodings_iso2022
- [ ] lib-python.3.test.test_multiprocessing_spawn
- [ ] lib-python.3.test.test_unicode

Additionally, on py3.10, maybe due to the way I updated the stdlib (only merged from py3.9, did not sync with cpython 3.10.14)
- [ ] lib-python.3.test.test_ensurepip
- [ ] lib-python.3.test.test_tarfile
- [ ] lib-python.3.test.test_tempfile
- [ ] lib-python.3.test.test_xml_etree
- [ ] lib-python.3.test.test_xml_etree_c

[...]

PR #4933, opened 2024-04-02
Vendor expat

mattip:
Update expat to 2.6 and vendor the sources into PyPy (like CPython) so it can statically link the library into libpypy. Also update `_ssl` for Python 3.9.19


mattip:
Related to #4927

PR #4935, opened 2024-04-05
update stdlib to python3.9.19

mattip:


hroncok:
Hey, when updating PyPy 3.9 to 7.3.16 I noticed the setuptools version goes backward and that the ensurepip module no longer works with WHEEL_PKG_DIR.

I can live with that, however, it was a bit surprising.

mattip:
Hmm. I guess that brought PyPy into line with CPython 3.9.19, which has setuptools-58.1.0. At some point earlier I must have updated it to setuptools-65.5.0 from CPython 3.10, and then this PR reverted that. Same goes for WHEEL_PKG_DIR: it must have been backported from 3.10, and then this PR reverted to the CPython 3.9.19 version of ensurepip which does not have WHEEL_PKG_DIR.

mattip:
I reapplied 8451984 in 82f8576ed3e.

hroncok:
Thanks.

Issue #4962, opened 2024-05-23
Some small missing items when compared to CPython


devdanzin:
These properties are missing from PyPy 3.10 on Windows, while present on CPython 3.10:

 - ~~array.typecodes~~ 
 - ~~cmath.tau~~
 - ~~sqlite3.complete_statement~~
 - ~~sqlite3.enable_callback_tracebacks~~ 
 - ~~sqlite3.SQLITE_CREATE_VTABLE~~
 - ~~sqlite3.SQLITE_DROP_VTABLE~~
 - ~~sqlite3.SQLITE_FUNCTION~~
 - ~~sqlite3.SQLITE_RECURSIVE~~
 - ~~sqlite3.SQLITE_SAVEPOINT~~
 - token.SOFT_KEYWORD
 - ~~ssl.OP_NO_RENEGOTIATION~~
 - ~~ssl.VERIFY_X509_PARTIAL_CHAIN~~
 - ~~stat.IO_REPARSE_TAG_APPEXECLINK~~
 - ~~stat.IO_REPARSE_TAG_MOUNT_POINT~~
 - ~~stat.IO_REPARSE_TAG_SYMLINK~~

Plus a bunch of `socket` and `zlib` constants that I can list if that would help.

mattip:
The `stat.IO_REPARSE*` constants were added in 1f46a199

mattip:
I added the _ssl and _sqlite3 ones. As for token.SOFT_KEYWORD, I am not sure why it is not exposed in pypy/interpreter/pyparser/pytoken.py. Maybe an oversight? I also see that lib-python/3/token.py exists. I think we should delete that, it is copied from CPython and is not relevant for PyPy.

mrdaybird:
@mattip I recently came across the issue regarding pypy's ssl module not having the `OP_NO_RENEGOTIATION` flag, nonetheless it's great to see that `ssl.OP_NO_RENEGOTIATION` is getting added to pypy. 
But do you have workaround to disabling renegotiation without this flag? FYI, RFC 9113 requires renegotiation to be disabled for HTTP/2 and it's a blocker for [urllib/3456](https://github.com/urllib3/urllib3/pull/3456)

mattip:
[...]

PR #4963, opened 2024-05-23
Add tau to cmath

devdanzin:
Attempt to add `cmath.tau`, aiming to address part of #4962.


devdanzin:
Hm, this PR is borked, should be two commits only. Sorry about that, I'll try again.

PR #4964, opened 2024-05-23
Add cmath.tau

devdanzin:
Use `math.pi * 2.0` to define it, as `math.tau` isn't available on Python 2.

Addresses part of #4962.


mattip:
Looks good, but I think this should target py3.9, and should also inject the constant into the `math` module

devdanzin:
Continued on #4971.

PR #4965, opened 2024-05-23
Add array.typecodes

devdanzin:
Copy `array.typecodes` from CPython, but assert that they match the set of typecodes used in PyPy code.

Addresses part of #4962.


mattip:
This should target py3.9

devdanzin:
Continued in #4970.

PR #4970, opened 2024-05-26
Add array.typecodes, take 2

devdanzin:
Copy `array.typecodes` from CPython, but assert that they match the set of typecodes used in PyPy code.

Hopefully targeting the correct branch this time :)

Addresses part of https://github.com/pypy/pypy/issues/4962.


cfbolz:
Looks good! Thank you

PR #4971, opened 2024-05-26
Add cmath.tau, take 2

devdanzin:
Use `math.pi * 2.0` to define it, as `math.tau` isn't available on Python 2.

@mattip said:
> should also inject the constant into the `math` module

But I don't know what that means, or how to do it. Any clues?

Addresses part of https://github.com/pypy/pypy/issues/4962.


mattip:
Sorry, my bad. `math.tau` already exists.

Issue #4992, opened 2024-07-29
Wrong value of "readonly" in PyMemoryView_FromMemory


bemoody:
When a C extension calls `PyMemoryView_FromMemory`, the resulting memoryview object has its "readonly" attribute backwards: if the caller specifies `PyBUF_READ`, the object is *not* read-only, and if the caller specifies `PyBUF_WRITE`, the object *is* read-only.

(CPython 3.11 behaves as documented here: https://docs.python.org/3.11/c-api/memoryview.html#c.PyBUF_READ)

I'm guessing that the bug is the following line (`pypy/module/cpyext/memoryobject.py`, line 197):
```
    readonly = int(widen(flags) == PyBUF_WRITE)
```
which should be instead:
```
    readonly = int(widen(flags) == PyBUF_READ)
```

(For the record, I'm aware that cpyext is not the most efficient way to interface C code with PyPy, and I'm aware that PyMemoryView_FromMemory is not the best API in general; but my goal for the moment is to have something that works and is portable across many Python versions.)

## Example

`exmodule.c`:
```
#include "Python.h"

static char buf[10];

static PyObject *
ex_mview_ro_memory(PyObject *self, PyObject *args)
{
    return PyMemoryView_FromMemory(buf, sizeof(buf), PyBUF_READ);
}

static PyObject *
ex_mview_rw_memory(PyObject *self, PyObject *args)
{
    return PyMemoryView_FromMemory(buf, sizeof(buf), PyBUF_WRITE);
}

static PyMethodDef ex_methods[] = {
    {"mview_ro_memory", ex_mview_ro_memory, METH_VARARGS, NULL},
    {"mview_rw_memory", ex_mview_rw_memory, METH_VARARGS, NULL},
    {NULL},
};

static struct PyModuleDef exmodule = {
    PyModuleDef_HEAD_INIT,
    "ex",
    NULL,
    0,
    ex_methods,
    NULL,
    NULL,
    NULL,
    NULL
};

PyMODINIT_FUNC
PyInit_ex(void)
{
    return PyModuleDef_Init(&exmodule);
}
```
`setup.py`:
```
from setuptools import Extension, setup

setup(
    name="ex",
    ext_modules=[
        Extension(
            name="ex",
            sources=["exmodule.c"],
        ),
    ],
)
```
`test.py`:
```
import sys
import ex
print(f'{sys.version=}')
print(f'{ex.mview_ro_memory().readonly=}')
[...]

PR #4993, opened 2024-07-29
Fix readonly flag in PyMemoryView_FromMemory

bemoody:
Fix issue #4992: the behavior of PyMemoryView_FromMemory is backwards.

**I haven't tested this**, but it seems right.

I'm also not sure of the procedure for making changes here.  I'm targetting this pull request at `py3.9` branch, but note that:

- The bug is also present in `py3.10` branch (as well as older versions.)

- This is an API break, so I don't know whether it's okay to change this in a stable release.

### DIFF
--- a/pypy/module/cpyext/memoryobject.py
+++ b/pypy/module/cpyext/memoryobject.py
@@ -194,7 +194,7 @@ def PyMemoryView_FromMemory(space, mem, size, flags):
     PyBUF_READ or PyBUF_WRITE. view->format is set to "B" (unsigned bytes).
     The memoryview has complete buffer information.
     """
-    readonly = int(widen(flags) == PyBUF_WRITE)
+    readonly = int(widen(flags) == PyBUF_READ)
     view = CPyBuffer(space, cts.cast('void*', mem), size, None,
             readonly=readonly)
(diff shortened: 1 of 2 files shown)


mattip:
Thanks. I assume the test fails before the fix, passes afterward. I guess that means not that many people are using that flag with PyPy in the wild.

We merge the py3.9 branch into py3.10, so the fix will percolate up.

It is not so much an API break as a bug fix and will be reported as such in the changelog when we release.

Is there a reason it is still marked Draft?

bemoody:
> Is there a reason it is still marked Draft?

I wanted to test and be sure I wasn't missing anything.  But I did and it seems to work.

<sub>(Love the build animation, by the way.)</sub>

mattip:
You can run tests without translation. This will run the new test on top of python2 untranslated, so you can add a python `import pdb;pdb.set_trace()` inside `pypy/module/cpyext/memoryobject.py` at the point of the change and explore the values, or let the test run to completion to see if it passes.
```
python2 pytest.py pypy/module/cpyext/test/test_memoryobject.py -k  test_FromMemory_readonly
```

mattip:
Thanks @bemoody

PR #5001, opened 2024-08-12
Fix typo

user-11150:


mattip:
Maybe `-m pip` is more correct, but on what platform does `-mpip` not work and could be considered a typo?

user-11150:
Sorry, I always thought it couldn't work without spaces, because everything I saw had spaces.

PR #5003, opened 2024-08-13
Use  the CPython3.13 _pyrepl module

mattip:
Works on linux and windows. There are some bits backported from python3.13, so this will require some care if we ever move to python3.11, python3.12.


mattip:
Thanks @cfbolz let's put this in.

Issue #4999, opened 2024-08-08
update cffi to 1.17


mattip:
CFFI 1.17 has been released. We should update for the next release.

mattip:
Also see #4937

PR #5008, opened 2024-08-21
update to cffi1.17.0

mattip:
- Run `PYTHONPATH=. python2 pypy/tools/release/import_cffi.py` to update extra_tests and lib_pypy
- Fix up lib_pypy/cffi.dist-info/METADATA
- Update pypy/module/_cffi_backend/test/_backend_test_c.py

The test_file test will still fail, since the cffi `main` branch did not update the version, see python-cffi/cffi#116

The METADATA will need re-doing when this is merged to py3.9/py3.10.


mattip:
closes #4999

PR #5014, opened 2024-08-30
Use Kyiv instead of Kiev

stefanor:
tzdata provides Kiev as an alternative to Kyiv, but Debian moved it to the tzdata-legacy package, breaking these tests.

cPython also uses this timezone in test_email, and this was fixed in 3.11. I've left that patch out of here, it will get picked up when PyPy switches to 3.11 stdlib.

https://github.com/python/cpython/pull/109251


mattip:
Thanks @stefanor

PR #5015, opened 2024-08-30
Prefer static sysconfigdata if it exists

stefanor:
We generate a static `_sysconfigdata` in our package, but we never use it.

In Debian we append some extra config into the static `_sysconfigdata` rather than patching it into `_sysconfigdata`. But that's no use unless the module is actually used.


mattip:
Thanks @stefanor

mattip:
I was too quick to merge this, I will have to revert. It will break portable builds, since it hard-codes the paths into the platform-specific `_sysconfigdata__*.py` file. We need a different solution. I added the file for the cross-compile package, which reads the `_sysconfig__*.py` file for compilation flags but if I recall correctly ignores the paths.

lib-python/3/sysconfig.py
```diff
@@ -493,9 +493,15 @@ def _init_posix(vars):
...
-    _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
+    #       generating them, fall back to the dynamic version here.
+    name = _get_sysconfigdata_name()
+    try:
+        _temp = __import__(name, globals(), locals(), ['build_time_vars'], 0)
+    except ImportError:
+        if '_PYTHON_SYSCONFIGDATA_NAME' in os.environ:
+            raise
+        _temp = __import__("_sysconfigdata", globals(), locals(), ['build_time_vars'], 0)
+
```
mattip, line 504: Maybe we could add a check here that the _temp['INCLUDEPY'] dir exists, and if not use the dynamic one?
```
    else:
        if not os.path.exists(_temp['INCLUDEPY']):
            # Use the dynamic module
            _temp = __import__("_sysconfigdata", globals(), locals(), ['build_time_vars'], 0) 

```

stefanor:
> It will break portable builds, since it hard-codes the paths into the platform-specific `_sysconfigdata__*.py` file.

I would suggest just not shipping the static file in the portable builds if it's not useful.

mattip:
OK, removing it. I originally added this for conda-forge, but they are dropping PyPy so indeed, maybe it is not needed anymore. If it is needed, the following shell script can be used to regenerate it (from the conda build, for future me when I need to revisit this topic):
```
# Regenerate the sysconfigdata__*.py file with paths from $PREFIX, at install
# those paths will be replaced with the actual user's paths. The generator
# builds the file in ./build/lib-<platform_tag>
host_gnu_type=$(${RELEASE_DIR}/config.guess)
pushd /tmp
pypy -m sysconfig --generate-posix-vars HOST_GNU_TYPE $host_gnu_type
cp build/*/_sysconfigdata*.py $PREFIX/lib/pypy${PY_VERSION}
popd
```

mattip:
disabled in 7802d077606.

PR #5022, opened 2024-09-03
fixup int/long confusion on 32-bit builds

mattip:
Goes a long way to addressing #5016 on main, but there is still a problem around `intptr_t` in the dump callback for the built-in tracback module. 

cc @stefanor


rpython/translator/platform/linux.py
```diff
@@ -14,7 +14,13 @@ class Linux(BasePosix):
...
     cflags = tuple(
              ['-O3', '-pthread', '-fomit-frame-pointer',
-              '-Wall', '-Wno-unused', '-Wno-address']
+              '-Wall', '-Wno-unused', '-Wno-address',
+              '-Wno-discarded-qualifiers',  # RPyField does not know about const
+              # The parser turns 'const char *const *includes' into 'const const char **includes'
+              '-Wno-duplicate-decl-specifier',
+              '-Werror=incompatible-pointer-types', '-Werror=implicit',
+              '-Werror=int-conversion',
+             ]
```
mattip, line 23: Lines 21-22 will simulate gcc-14 error checking on older gcc.

pypy/module/_multiprocessing/interp_semaphore.py
```diff
@@ -202,8 +208,9 @@ def sem_getvalue(sem):
 
     def gettimeofday():
         now = lltype.malloc(TIMEVALP.TO, 1, flavor='raw')
+        void = lltype.nullptr(rffi.VOIDP.TO)
         try:
-            res = _gettimeofday(now, None)
+            res = _gettimeofday(now, void)
```
mattip, line 213: This should use c_gettimeofday from rpython/rlib/rtime

pypy/module/faulthandler/cintf.py
```diff
@@ -26,7 +26,7 @@ def direct_llexternal(*args, **kwargs):
 
 
 DUMP_CALLBACK = lltype.Ptr(lltype.FuncType(
-                     [rffi.INT, rffi.SIGNEDP, lltype.Signed], lltype.Void))
+                     [rffi.INT_real, rffi.INTPTR_TP, rffi.INTPTR_T], lltype.Void))
```
mattip, line 29: The callback signature uses `intptr_t *`, but that still is not working here.

stefanor: Yeah, I see:
```
	pypy_module_faulthandler.c: In function 'pypy_g_Handler_enable':
	pypy_module_faulthandler.c:117:45: error: passing argument 1 of 'pypy_faulthandler_setup' from incompatible pointer type [-Wincompatible-pointer-types]
	  117 |         l_v215550 = pypy_faulthandler_setup(pypy_g__dump_callback);
	      |                                             ^~~~~~~~~~~~~~~~~~~~~
	      |                                             |
	      |                                             void (*)(int,  Signed *, Signed) {aka void (*)(int,  long int *, long int)}
	In file included from singleheader.h:141:
	/home/stefanor/pypy/pypy/module/faulthandler/faulthandler.h:11:65: note: expected 'pypy_faulthandler_cb_t' {aka 'void (*)(int,  int *, int)'} but argument is of type 'void (*)(int,  Signed *, Signed)' {aka 'void (*)(int,  long int *, long int)'}
	   11 | RPY_EXTERN char *pypy_faulthandler_setup(pypy_faulthandler_cb_t dump_callback);
	      |                                          ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
	pypy_module_faulthandler.c: In function 'pypy_g_Handler_dump_traceback':
```

pypy/module/_multiprocessing/interp_semaphore.py
stefanor: This also brings up 64-bit time
Glibc doc: https://sourceware.org/glibc/wiki/Y2038ProofnessDesign

Debian uses 64-bit time on all 32-bit architectures except i386 (i386 mostly exists for compatibility for older binaries)

mattip:
I think part (most?) of this was caused by the [changes made for win64](https://doc.pypy.org/en/latest/windows64.html?highlight=signed#what-we-did)
> This runs small translations, and some details were wrong. The most obvious one is to make the integer type `Signed` equal to `long` on every other platform, but on Win64 it should be something like `long long`.

On 32-bit i686, pointers are 32-bits, so Signed should default to `int` not `long`. Now if I could only find if that hypothesis is correct and there was some change to the way `Signed` is converted to `C` that subtly broke 32-bit i686.

stefanor:
That's translating on armhf, now. Just faulthandler remaining as a translation-blocker.

mattip:
Locally, this now translates on i686 and x86_64. I removed the cflags since they made a test fail. @stefanor let me know what you think.

pypy/module/faulthandler/cintf.py
```diff
@@ -26,7 +26,7 @@ def direct_llexternal(*args, **kwargs):
 
 
 DUMP_CALLBACK = lltype.Ptr(lltype.FuncType(
-                     [rffi.INT, rffi.SIGNEDP, lltype.Signed], lltype.Void))
+                     [rffi.INT_real, rffi.VOIDP, rffi.INT_real], lltype.Void))
```
mattip, line 29: What we really want here is a platform `intptr_t`, but it seems hard to get the lltype.Number and genc to actually output `intptr_t` instead of `Signed`.

stefanor:
> I wonder why they chose to emit Signed for so many int types, and didn't emit the C name like intptr_t instead.

I assume that was just what made sense for Windows's LLP64 64-bit model. And nobody put too much thought into the LP64 interfaces that were breaking, because it's all little endian and just worked.

stefanor (approved):
LGTM. Translates on amd64, i386 and armhf

mattip:
Fixes #5016

mattip:
Someday we should refactor all this and emit the `c_name` instead of `Signed` by default.

pypy/module/select/interp_epoll.py
```diff
@@ -190,16 +190,18 @@ def descr_poll(self, space, timeout=-1.0, maxevents=-1):
...
-            with lltype.scoped_alloc(rffi.CArray(rffi.INT), maxevents) as events:
+        with lltype.scoped_alloc(rffi.CArray(rffi.UINT_real), maxevents) as fids:
+            with lltype.scoped_alloc(rffi.CArray(rffi.INT_real), maxevents) as events:
                 nfds = pypy_epoll_wait(self.epfd, fids, events, maxevents, int(timeout))
                 if nfds < 0:
                     raise exception_from_saved_errno(space, space.w_IOError)
 
                 elist_w = [None] * nfds
                 for i in xrange(nfds):
+                    fid = rffi.cast(rffi.UINT_real, fids[i])
```
stefanor, line 201: unused variable

PR #5024, opened 2024-09-03
Fix socket.socket.sendto for AF_PACKET protocol

tecki:
the hardware address for an AF_PACKET socket is 8 bytes, not a string.

This is how it is in CPython, where a PyArg_ParseTuple with data type y* is used.

### DIFF
--- a/pypy/module/_socket/interp_socket.py
+++ b/pypy/module/_socket/interp_socket.py
@@ -120,7 +120,7 @@ def addr_from_object(family, fd, space, w_address):
         else:                 pkttype = 0
         if len(pieces_w) > 3: hatype = space.int_w(pieces_w[3])
         else:                 hatype = 0
-        if len(pieces_w) > 4: haddr = space.text_w(pieces_w[4])
+        if len(pieces_w) > 4: haddr = space.bytes_w(pieces_w[4])
         else:                 haddr = ""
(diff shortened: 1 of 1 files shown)


mattip:
Makes sense. That came in from 03fff1a (in 2015!), which added [AppTestPacket](https://github.com/pypy/pypy/blob/62eb9b341e28a581dd41ee4fdf11722479a86939/pypy/module/_socket/test/test_sock_app.py#L710) to test the parsing. Can you extend those tests to reach this code? You can run the tests untranslated with `python2 pytest.py pypy/module/_socket/test/test_sock_app.py -k AppTestPacket`

tecki:
I would love to, I just don't know how to. The problem is that this bug does not exist in Python 2, as it does not make a difference between strings and bytes, so the problem shows up only in Python 3 (probably this is why this problem has not been seen before). The tests, however, are written in Python 2.

mattip:
Ahh, and the test is decorated with this, so it is skipped unless run as root. I guess that is why we did not see it.
```
@pytest.mark.skipif(not hasattr(os, 'getuid') or os.getuid() != 0,
    reason="AF_PACKET needs to be root for testing")
```

tecki:
Thanks for merging!

PR #5032, opened 2024-09-17
Py3.10 utf8 rposix

### LINKED ISSUE #5000: Windows: PyPy Interpreter raises ModuleNotFoundError if path to interpreter contains Umlaut
As the title suggests, we run into a `ModuleNotFoundError` when the path to the interpreter contains an Umlaut or probably any non-ascii characters:

```
$ /d/python/ü/pypy3.10-v7.3.16-win64/pypy.exe
ModuleNotFoundError: No module named 'encodings'
debug: OperationError:
debug:  operror-type: ModuleNotFoundError
debug:  operror-value: No module named 'encodings'
```

- OS: Windows-10
- PyPy version: pypy3.10-v7.3.16-win64 (I have also tested older pypy versions with the same result)
- Installation: Extracted PyPy via [pypy3.10-v7.3.16-win64.zip](https://downloads.python.org/pypy/pypy3.10-v7.3.16-win64.zip)
- Env: My environment does not contain any Python related environment variables

Please let me know if there is more information needed.

Thanks in advance!

mattip:
Fixes #5000, mainly by using wchar interfaces when calling windows API functions in initpath.py and `rpython dlopen()`, but also cleans out the use of unicode when calling rpython.


pypy/interpreter/app_main.py
```diff
@@ -393,6 +394,8 @@ def initstdio(encoding=None, unbuffered=False):
         else:
             errors = None
         encoding = encoding or None
+        if _WIN32 and not encoding:
+            encoding = "utf-8"
```
mattip, line 398: I thought this might fix the console use of unicode, but it didn't help.

pypy/interpreter/unicodehelper.py
```diff
@@ -68,9 +68,8 @@ def fsdecode(space, w_string):
     if rutf8.first_non_ascii_char(utf8) < 0:
         return space.newtext(utf8, len(utf8))
     if _WIN32:
-        import pypy.interpreter.unicodehelper_win32 as win32
         slen = len(utf8)
-        utf8, _, lgt = str_decode_utf8(space, utf8, w_string, 'surrogateescape', True, errorhandler)
+        utf8, lgt, _ = str_decode_utf8(space, utf8, w_string, 'surrogateescape', True, errorhandler)
```
mattip, line 72: This was causing lots of problems. It created invalid W_Unicode objects by using the utf8 length not the codepoints one.

Issue #5033, opened 2024-09-17
thread switching in windows is slow


mattip:
Windows runs the [`test_ping_pong` test](https://github.com/pypy/pypy/blob/058ccb4d58ad592903747203a70a261d8ad00d5c/pypy/module/thread/test/test_lock.py#L91) much slower than posix. I think the problem is that we are using a very old version of [`thread_nt.c`](https://github.com/pypy/pypy/blob/main/rpython/translator/c/src/thread_nt.c), where CPython moved to one that has [this at the top](https://github.com/python/cpython/blob/33eeccf6d4f16e483b4c8a180bad718545aeaeaf/Python/thread_nt.h#L19):
```
/* Now, define a non-recursive mutex using either condition variables
 * and critical sections (fast) or using operating system mutexes
 * (slow)
 */
```

It seems our version is using the operating system mutexes. This was changed in 2012 for CPython 2.7.

mattip:
I wonder if that has anything to do with windows being over 2x slower. Here is the timing from a translation with pypy2.7-v7.3.14
```
[Timer] Timings:
[Timer] loading target                 ---   79.3 s
[Timer] annotate                       ---  531.8 s
[Timer] rtype_lltype                   ---  252.7 s
[Timer] pyjitpl_lltype                 ---  519.3 s
[Timer] backendopt_lltype              ---  189.2 s
[Timer] stackcheckinsertion_lltype     ---    8.4 s
[Timer] database_c                     ---  353.2 s
[Timer] source_c                       ---  217.2 s
[Timer] compile_c                      ---  107.2 s
[Timer] build_cffi_imports             ---   28.6 s
[Timer] ===========================================
[Timer] Total:                         --- 2287.0 s
```

vs. a translation on the same machine running Ubuntu, with the same 2.7-v7.3.14
```
[Timer] Timings:
[Timer] loading target                 ---  23.2 s
[Timer] annotate                       --- 216.3 s
[Timer] rtype_lltype                   ---  96.8 s
[Timer] pyjitpl_lltype                 --- 205.1 s
[Timer] backendopt_lltype              ---  82.6 s
[Timer] stackcheckinsertion_lltype     ---   7.1 s
[Timer] database_c                     --- 157.4 s
[Timer] source_c                       ---  73.1 s
[Timer] compile_c                      ---  53.0 s
[Timer] build_cffi_imports             ---   3.0 s
[Timer] ==========================================
[Timer] Total:                         --- 917.5 s
```

[...]

PR #5063, opened 2024-10-02
refactor windows gil mutex handling to mimic pthreads

mattip:
Now that the threading primitives match more closely, use these in a similar way to the pthreads code. Maybe solves #5033 (after merging to py3.10)

### DIFF
--- a/rpython/translator/c/src/thread_win7.c
+++ b/rpython/translator/c/src/thread_win7.c
@@ -41,81 +41,6 @@ typedef SRWLOCK PyMUTEX_T;
 typedef CONDITION_VARIABLE  PyCOND_T;
 
 /* -------------------------
- * From Python/condvar.h
- */
-
-
-Py_LOCAL_INLINE(int)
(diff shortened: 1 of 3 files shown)


mattip:
Using this version, here is the timing for a windows translation of `main`. No large change from #5056 ☹️ 
```
[Timer] Timings:
[Timer] loading target                 ---   30.2 s
[Timer] annotate                       ---  355.8 s
[Timer] rtype_lltype                   ---  208.2 s
[Timer] pyjitpl_lltype                 ---  444.3 s
[Timer] backendopt_lltype              ---  135.4 s
[Timer] stackcheckinsertion_lltype     ---    5.9 s
[Timer] database_c                     ---  235.2 s
[Timer] source_c                       ---  133.3 s
[Timer] compile_c                      ---   68.5 s
[Timer] build_cffi_imports             ---    9.9 s
[Timer] ===========================================
[Timer] Total:                         --- 1627.0 s
```

Issue #5079, opened 2024-10-15
Race in `Thread.join()`


x42005e1f:
Hello, I was thinking and decided that I should let you know about a problem with the implementation of the [threading](https://docs.python.org/3/library/threading.html) module. As we all know, PyPy implements the [_thread](https://docs.python.org/3/library/_thread.html) module, but leaves the [threading](https://docs.python.org/3/library/threading.html) module as it is in CPython. As a consequence, PyPy suffers from [CPython's gh-116372](https://github.com/python/cpython/issues/116372). This leads to [RuntimeError](https://github.com/microsoft/vscode-python-debugger/issues/256), [AssertionError](https://github.com/pypy/pypy/issues/3474), and even [SEGFAULT](https://github.com/pypy/pypy/issues/2531).

In [my package](https://github.com/x42005e1f/aiologic), which is directly related to threading, I have implemented a [monkey patch](https://github.com/x42005e1f/aiologic/blob/760ab3dbe9cb59d1f3bb3dcd90b5a85b1f6f639c/src/aiologic/lowlevel/patcher.py#L403) that fixes this issue. You can also see a more detailed description of the problem in the function description. Hopefully, it will help you better understand the nature of the many related reports of threading issues that remain unresolved. Because, as I can see from [another issue](https://github.com/microsoft/debugpy/issues/1198), this problem is still shrouded in mystery.

mattip:
Thanks. Would you like to submit a PR to target the py3.10 branch?

mattip:
Maybe related to the timeouts on the [`test_del_shutdown (test.test_concurrent_futures.ThreadPoolShutdownTest)`](https://buildbot.pypy.org/summary/longrepr?testname=unmodified&builder=pypy-c-jit-linux-x86-64&build=11224&mod=lib-python.3.test.test_concurrent_futures) test on the buildbot

mattip:
Thanks, committed in c9795377602

PR #5080, opened 2024-10-15
Fix race in Thread.join()

x42005e1f:
This is a very simple PR that fixes [the race](https://github.com/pypy/pypy/issues/5079) by removing one problematic `assert` statement and suppressing `RuntimeError` raised due to multiple `lock.release()` calls. If it seems strange to some, well, it is. I too wanted a more sophisticated solution.

More in-depth details are in the [my patch description](https://github.com/x42005e1f/aiologic/blob/760ab3dbe9cb59d1f3bb3dcd90b5a85b1f6f639c/src/aiologic/lowlevel/patcher.py#L403). There are also tests you can use to check the solution. And unlike my patch, it doesn't make additional optimizations as they are not relevant to the solution.

Side effect: direct `Thread._stop()` calls will succeed. But it is quite difficult to distinguish such from a typical race condition, so the lesser evil is chosen.

### DIFF
--- a/lib-python/3/threading.py
+++ b/lib-python/3/threading.py
@@ -1043,8 +1043,9 @@ class Thread:
         # Special case:  _main_thread releases ._tstate_lock via this
         # module's _shutdown() function.
         lock = self._tstate_lock
-        if lock is not None:
-            assert not lock.locked()
+        # XXX PyPy change: remove assert statement that fails at race condition
+        # if lock is not None:
+        #     assert not lock.locked()
         self._is_stopped = True
         self._tstate_lock = None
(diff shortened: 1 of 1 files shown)


mattip:
>  If it seems strange to some, well, it is. I too wanted a more sophisticated solution

Indeed, it does seem strange to ignore failed calls to `lock.release()`, but all this is at thread shutdown so :shrug:. Is this related to debugpy/issues/#1198?

mattip:
Thanks @x42005e1f. Let's see if this fixes the timeouts in test_contextlib_async

PR #5085, opened 2024-10-18
avoid keeping a ref to scope_w values while running self.behaviour

### LINKED ISSUE #5075: gc.get_referrers returns an extra redundant tuple
```python
import gc

class Ham:
    pass

def main():
    h = Ham()
    print(gc.get_referrers(h))

main()
```

prints:

```sh
pypy3 demo.py                      
[<frame at 0x00007be407e56020, file '/home/graingert/projects/anyio/demo.py', line 8, code main>, (<__main__.Ham object at 0x00007be407bbf440>,)]
```

I think the extra tuple is this `args_w` here: https://github.com/pypy/pypy/blob/72e9257dde455ca17884b409b62c9088508b633f/pypy/module/gc/referents.py#L134

I think it should be dropped from the returned list

graingert:
Fixes #5075

### DIFF
--- a/pypy/interpreter/gateway.py
+++ b/pypy/interpreter/gateway.py
@@ -249,7 +249,7 @@ class UnwrapSpec_EmitRun(UnwrapSpecEmit):
         self.run_args = []
 
     def scopenext(self):
-        return "scope_w[%d]" % self.succ()
+        return "self._move_from_scope(scope_w, %d)" % self.succ()
 
     def visit_self(self, typ):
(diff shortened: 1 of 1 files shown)


mattip:
Is there a test for this change?

graingert:
I'll be trying another approach shortly and will write a test

mattip:
Any progress here? Need any help?

graingert:
I don't think this approach is viable, and I'm distracted looking at other things. Will get back to this bug soon

PR #5087, opened 2024-10-20
avoid keeping refs on the frame stack when calling functions

graingert:
This fixes the case where the following program hangs in pypy3.10 and terminates in cpython 3.11:

```python
import sys
import gc
import weakref

class A:
    pass

def other(a):
    v = []  # using a list rather than nonlocal for pypy2 support

    # delete the only remaining ref to a and replace with a weakref
    a = weakref.ref(a, lambda _: v.append(True))
    while not v:
        # on cpython this branch is never hit, but on pypy we need to sit and
        # collect, until the object is deleted
        gc.collect()


def main():
    # move our reference to the instance of A into other
    other(A())

if __name__ == "__main__":
    sys.exit(main())
```


graingert:
I'd like to know how this impacts performance, and then I'll write a test for it

mattip:
> I'd like to know how this impacts performance

Are you asking for help or stating that you yourself will check performance and get back to this?

graingert:
@mattip I'm asking for help on this

mattip:
So we can run the benchmarks by pushing the branch to this repo, then kicking off [the buildbot](https://buildbot.pypy.org/builders/jit-benchmark-linux-x86-64) on the branch. Once they finish, the branch should show up on the [benchmark comparison](https://speed.pypy.org/comparison/) page. I will do that.

mattip:
The benchmarks [are running](https://buildbot.pypy.org/builders/jit-benchmark-linux-x86-64/builds/4257), typically it takes ~2 hours to finish.

mattip:
Some [untranslated tests are failing](https://buildbot.pypy.org/summary?builder=own-linux-x86-64&branch=avoid-keeping-refs-when-calling-fns), they pass on `main`.

graingert:
how do I see the failure tracebacks? and how do I run them locally?

mattip:
You can click on the `F` or the `!` to see the traceback. You can run them locally with `pypy2.7 pytest.py <testfile>` -k <testname>`, so the first crash can be run locally with
```
pypy2.7 pytest.py pypy/objspace/std/mapdict/py 
```

mattip:
Gahh. Benchmarking failed to upload the results because `len(branch name)==35` but the [model](https://github.com/python/codespeed/blob/263860bc298fd970c8466b3161de386582e4f801/codespeed/migrations/0003_project_default_branch.py#L32) only allows 32.

mattip:
I uploaded the benchmark results as the branch "avoid-keeping-refs". You can see in [the comparison](https://speed.pypy.org/comparison/) that the branch does not seem to change things much. Some benchmarks are slower, some faster. How to see it? For runs with `pypy --jit off` do:
- Environments: check only "benchmarker"
- Executables: check only "pypy-64 latest" and "pypy-64 latest in branch 'avoid-keeping-refs'"
- Normalization: pypy-64-latest
- Horizontal view

For the jitted benchmarks:
- Environments: check only "benchmarker"
- Executables: check only "pypy-jit-64 latest" and "pypy-jit-64 latest in branch 'avoid-keeping-refs'"
- Normalization: pypy-jit-64-latest
- Horizontal view

graingert:
> You can click on the `F` or the `!` to see the traceback. You can run them locally with `pypy2.7 pytest.py <testfile>` -k `, so the first crash can be run locally with
> 
> ```
> pypy2.7 pytest.py pypy/objspace/std/mapdict/py 
> ```

this doesn't seem to do anything on my local checkout:

```
 ✘  graingert@conscientious  pypy2  ~/projects/pypy   avoid-keeping-refs-when-calling-fns ± python pytest.py pypy/objspace/std/mapdict/py 
======================================================================= test session starts ========================================================================
platform linux2 -- Python 2.7.18[pypy-7.3.17-final], pytest-2.9.2, py-1.4.29, pluggy-0.3.1
rootdir: /home/graingert/projects/pypy, inifile: pytest.ini

=================================================================== no tests ran in 0.00 seconds ===================================================================
ERROR: file not found: pypy/objspace/std/mapdict/py
 ✘  graingert@conscientious  pypy2  ~/projects/pypy   avoid-keeping-refs-when-calling-fns ± python pytest.py pypy/objspace/std/mapdict.py 
======================================================================= test session starts ========================================================================
platform linux2 -- Python 2.7.18[pypy-7.3.17-final], pytest-2.9.2, py-1.4.29, pluggy-0.3.1
pytest-2.9.2 from /home/graingert/projects/pypy/pytest.pyc
rootdir: /home/graingert/projects/pypy, inifile: pytest.ini
collected 0 items 

=================================================================== no tests ran in 0.82 seconds ===================================================================
```

mattip:
Sorry, typo. The general pattern is `<python2> pytest.py <filename>`. I got the filename wrong, the suffix is `.py` not `/py` so
```
pypy2.7 pytest.py pypy/objspace/std/mapdict.py 
```

graingert:
I tried both and the second run `collected 0 items `

mattip:
This will translate enough of PyPy to run the tests in the file. In this case, it will translate the basic interpreter without any optional modules. The first time it is run it caches a bunch of stuff in `rpython/_cache`, mostly about the host platform, so subsequent test runs should be a little faster.

mattip:
Hmm. So what is [failing on the buildbot](https://buildbot.pypy.org/summary?branch=avoid-keeping-refs-when-calling-fns)?

cfbolz:
The filename is still wrong ;-)

mapdict.py is the implementation file, not the test file

mattip:
Gahh. So the tests to fix or skip are in `pypy/module/_continuation/test/test_zpickle.py`

cfbolz:
@graingert sorry, I have tons of stuff this week and can't look super carefully. what we will definitely need to do, though, is to write some microbenchmarks that call small functions with varying amounts of arguments and see whether anything changes. I would guess/hope not, the interpreter is doing the same amount of work before and after the change it seems to me, but we should still check. Those should be run without the JIT, since the change only affects the interpreter anyway.

re the tests in `pypy/module/_continuation/test/test_zpickle.py`: that's very much a "here be dragons" file/directory. I'm happy to take a look at those when everything is looking pretty good.

graingert:
@cfbolz no worries I've found it's possible to work around in most cases by using a keyword argument:

eg

```python
import sys
import gc
import weakref

class A:
    pass

def other(a):
    v = []  # using a list rather than nonlocal for pypy2 support

    # delete the only remaining ref to a and replace with a weakref
    a = weakref.ref(a, lambda _: v.append(True))
    while not v:
        # on cpython this branch is never hit, but on pypy we need to sit and
        # collect, until the object is deleted
        gc.collect()


def main():
    # move our reference to the instance of A into other
    other(a=A())

if __name__ == "__main__":
    sys.exit(main())
 ```

I've pushed a fix for zpickle and I'd like to get another test run?

File: pypy/interpreter/baseobjspace.py [hunk 1]
```diff
@@ -1239,10 +1240,12 @@ def call_valuestack(self, w_func, nargs, frame, methodcall=False):
 
             if isinstance(w_func, Function):
                 return w_func.funccall_valuestack(
-                        nargs, frame, methodcall=methodcall)
+                        nargs, frame, methodcall=methodcall, dropvalues=dropvalues)
+
```
mattip, line 1244: Here and elsewhere: dropvalues is not really a kwarg

graingert: I don't understand, does that cause a problem?

graingert:
it passed, now to write some tests

cfbolz:
@graingert this looks good to me now. should I merge it or was there anything you still wanted to change?

graingert:
I think it's ready for merge! We did discuss benchmarks for this as it's in a fast path

cfbolz:
I did some microbenchmarks on this a while ago, they were fine (as in, no difference).

graingert:
Oh great, can I merge it? Do you want a squash merge or particular message?

cfbolz:
feel free to merge! And no, we don't squash.

graingert:
Ah nice, did you know you can disable squash merges in the GitHub settings?

PR #5097, opened 2024-11-01
Fix test_concurrent_futures.py hangs

x42005e1f:
This PR fixes hangs in tests due to PyPy garbage collector behavior. Related to #4994.


mattip:
Thanks @x42005e1f

PR #5101, opened 2024-11-04
differentiate between rffi.UNSIGNED/rffi.SIGNED and rffi.SSIZE_T/rffi.SIZE_T

mattip:
The goal is to have different low-level types for rffi.SIZE_T (which is currently rffi.ULONG/rffi.UNSIGNED and rffi.SSIZE_T, that will become `size_t` and `ssize_t` when translated.

The first commit creates the separate rffi types. Now a mechanism is needed, like for Signed and others, to cast rpython arguments to them, like in the calls in _multibytecodec/c_codecs.py

The work has mainly been done for rffi.SIZE_T, more is probably needed for rffi.SSIZE_T

### DIFF
--- a/pypy/module/_cffi_backend/test/test_file.py
+++ b/pypy/module/_cffi_backend/test/test_file.py
@@ -1,7 +1,6 @@
 import urllib2, py
 import pytest
 
-
 @pytest.mark.skip("cffi has moved onto v2.0, the PyPy backend is 1.18")
 def test_same_file():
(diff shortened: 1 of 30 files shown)


mattip:
Bleh. On 32-bit linux, `rffi.sizeof(lltype.Signed)` is 4, but `rffi.sizeof(lltype.SSize_T)` is 8. So they are not interchangeable, casting between them needs to check for overflow.

PR #5114, opened 2024-11-12
VMProf: Resolve names of native functions

Cskorpion:
use vmp_resolve_addr function to get name, file and line number of native functions after profiling with vmprof on pypy.

### DIFF
--- a/pypy/module/_vmprof/moduledef.py
+++ b/pypy/module/_vmprof/moduledef.py
@@ -22,7 +22,7 @@ class Module(MixedModule):
         'VMProfError': 'space.fromcache(interp_vmprof.Cache).w_VMProfError',
     }
     if rvmprof.supports_native_profiling():
-        interpleveldefs.update({'resolve_addr': 'interp_vmprof.vmprof_resolve_address'})
+        interpleveldefs['resolve_addr']= 'interp_vmprof.vmprof_resolve_address'
 
 
 # Force the __extend__ hacks and method replacements to occur
--- a/pypy/module/_vmprof/moduledef.py
+++ b/pypy/module/_vmprof/moduledef.py
@@ -17,6 +17,7 @@ class Module(MixedModule):
         'get_profile_path': 'interp_vmprof.get_profile_path',
         'stop_sampling': 'interp_vmprof.stop_sampling',
         'start_sampling': 'interp_vmprof.start_sampling',
+        'resolve_addr': 'interp_vmprof.vmprof_resolve_address',
 
         'VMProfError': 'space.fromcache(interp_vmprof.Cache).w_VMProfError',
     }
--- a/pypy/module/_vmprof/moduledef.py
+++ b/pypy/module/_vmprof/moduledef.py
@@ -1,5 +1,6 @@
 from pypy.interpreter.mixedmodule import MixedModule
 from rpython.rlib.rvmprof import VMProfPlatformUnsupported
+from rpython.rlib import rvmprof
 from rpython.translator.platform import CompilationError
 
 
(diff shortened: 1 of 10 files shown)


rpython/rlib/rvmprof/rvmprof.py
```diff
@@ -186,6 +186,34 @@ def start_sampling(self):
...
+                    namebuffer.raw[0] = '\x00'
+                    sourcefilebuffer.raw[0] = '-'
+                    sourcefilebuffer.raw[1] = '\x00'
+                    intbuffer = rffi.cast(rffi.INT_realP, intbuffer.raw)
+                    intbuffer[0] = rffi.cast(rffi.INT_real, 0)
+                    length = rffi.cast(rffi.INT, 256)
+                    res = self.cintf.vmprof_resolve_address(rffi.cast(rffi.VOIDP, addr), namebuffer.raw, length, intbuffer, sourcefilebuffer.raw, length)
+        
+                    #if res != 0:
+                    #    return (None, rffi.cast(rffi.SIGNED, -1), None)
```
cfbolz, line 212: what about this comment? can `vmprof_resolve_address` return something `!= 0`?

pypy/module/_vmprof/moduledef.py
```diff
@@ -20,6 +21,8 @@ class Module(MixedModule):
 
         'VMProfError': 'space.fromcache(interp_vmprof.Cache).w_VMProfError',
     }
+    if rvmprof.supports_native_profiling():
+        interpleveldefs.update({'resolve_addr': 'interp_vmprof.vmprof_resolve_address'})
```
cfbolz, line 25: you could just write `interpleveldefs['resolve_addr'] = 'interp_vmprof.vmprof_resolve_address'` here

rpython/rlib/rvmprof/cintf.py
```diff
@@ -20,12 +20,14 @@ class VMProfPlatformUnsupported(Exception):
...
 # vmprof works only on x86 for now
 IS_SUPPORTED = False
+NATIVE_PROFILING_SUPPORTED = False
 if sys.platform in ('darwin', 'linux', 'linux2') or sys.platform.startswith('freebsd'):
     try:
         proc = detect_cpu.autodetect()
         IS_SUPPORTED = (proc.startswith('x86')
                         or proc == 'aarch64'
                         or proc == 'riscv64')
+        NATIVE_PROFILING_SUPPORTED = proc.startswith('x86')# or proc.startswith('ppc-64')
```
cfbolz, line 30: what about the commented out part? I suspect it should just be removed?

Cskorpion: Native profiling seems to be supported on powerpc https://github.com/pypy/pypy/blob/f67cb822600ddfc00e4fa72b3ef8251c386087cb/rpython/rlib/rvmprof/src/shared/vmprof.h#L51-L54

But it also seems like vmprof is not supported on powerpc in general.
So yes, it can be removed

rpython/rlib/rvmprof/rvmprof.py
Cskorpion: yes, if `vmprof_resolve_address` could not find anything it returns `1`.
In that case we can just return the buffers with their default values.
But i guess explicitly handling that, would be better.

cfbolz:
looks good! will merge when the buildbots pass

mattip:
It might be nice to do a little "before and after" blog post that shows how the names now show up.

Cskorpion:
Thanks for the Review.
Are you referring to how `dladdr`, `lib-backtrace` and `mach` work and how `vmp_resolve_addr` uses them to retrieve information about native functions, or what the differences are in the 'frontend' of vmprof, like showing some example profiling with native names resolved?

cfbolz:
> what the differences are in the 'frontend' of vmprof, like showing some example profiling with native names resolved?

yes, I think that's what @mattip meant. it could be a super short post showing vmprof with native profiling before the change (only addresses are visible) and after the change (some addresses have names).

Cskorpion:
Sure, I'll do that.

mattip:
Thanks! This is a long-needed improvement

Issue #5133, opened 2024-11-23
implement a remote debugging mechanism


cfbolz:
I was talking with @pablogsal yesterday, and he described his idea for building a mechanism into a Python VM that would make it easier/possible to start a remote debugging session in an arbitrary Python process, using [`process_vm_writev`](https://linux.die.net/man/2/process_vm_writev).

I'll try to implement a prototype of this in PyPy, it's a very cool feature that I've wanted for ages.

mattip:
There was once upon a time [pyrasite](https://github.com/lmacken/pyrasite) for code injection and [bytehook](https://github.com/tzickel/bytehook) for adding a function hook. This could be combined with something like [rdb](https://py-toolbox.readthedocs.io/en/latest/modules/rdb.html#remote-debugging) which allows debugging over a socket.

cfbolz:
Oh, pyrasite is neat, I wasn't aware, thanks for the pointer! Yeah, it would be a lot like that, just implemented differently.

mattip:
Closing since the feature was implemented, please reopen or open a new issue if there is more to do here

PR #5135, opened 2024-11-25
make it possible to remotely inject python code into another python process

cfbolz:
see #5133 

this is the low level mechanism so far, it can be used to inject Python code into another PyPy process. Linux only, and you need root rights (or the ptrace settings need to be quite permissive).

### DIFF
--- a/pypy/doc/remotedebugging.rst
+++ b/pypy/doc/remotedebugging.rst
@@ -1,0 +1,31 @@
+Remote Debugging
+=================
+
+PyPy has an experimental remote debugging facility built into the virtual
+machine. It can be used to inject arbitrary Python code into another PyPy
+process. So far, it works only under Linux.
+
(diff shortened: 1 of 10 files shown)


pypy/doc/remotedebugging.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Security
+---------
+
+This feature is implemented by writing to the heap memory of another process.
+Under Linux, it uses the ``process_vm_writev`` function. This function is
+disabled by default on most hardened Linux distributions by Yama__. Therefore
+you might need sudo to run the above commands.
+
+.. __Yama: https://www.kernel.org/doc/html/v4.15/admin-guide/LSM/Yama.html#ptrace-scope
```
merwok, line 44: The link doesn’t work: two underscores is for self-contained inline link, one underscore is needed for a forward reference.

PR #5157, opened 2024-12-25
Add PyThread_acquire_lock_timed

da-woods:
It looks to map fairly directly onto RPyThreadAcquireLockTimed anyway so no new functionality is needed.

### DIFF
--- a/pypy/module/cpyext/api.py
+++ b/pypy/module/cpyext/api.py
@@ -633,7 +633,7 @@ SYMBOLS_C = [
     'PyOS_getsig', 'PyOS_setsig', 'PyType_GetName', 'PyType_GetQualName',
     '_Py_RestoreSignals',
     'PyThread_get_thread_ident', 'PyThread_allocate_lock', 'PyThread_free_lock',
-    'PyThread_acquire_lock', 'PyThread_release_lock',
+    'PyThread_acquire_lock', 'PyThread_acquire_lock_timed', 'PyThread_release_lock',
     'PyThread_create_key', 'PyThread_delete_key', 'PyThread_set_key_value',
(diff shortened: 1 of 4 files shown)


da-woods:
It looks like this was added to CPython in 3.2. I can't see anything else that's conditionally included by version, so I assume it's OK to define generally though.

mattip:
Is there a reason you targeted `main` which is Python2.7? We don't really intend on supporting the cpyext C-API compatibility layer there. 

This is a nice start, but it needs 
- a test
- to be reflected into `api.py`
- a fix, there is a [compilation failure](https://github.com/pypy/pypy/actions/runs/12492620865/job/34861497492?pr=5157#step:6:1445)

You can use `PyThread_acquire_lock` as an example for the first two problems.

da-woods:
Thanks. Is "Py3.11" the current branch I should be targeting?

> * a test

I'd initially only found `_testcapimodule.c` for tests (which I didn't want to modify because I think it's a direct copy from CPython) but I think I now see where the real tests are.

mattip:
> Thanks. Is "Py3.11" the current branch I should be targeting?

If this is a bug fix you can target py3.10, and we will merge it into py3.11 later.

To run tests without translating, we do
```
python2 pytest.py pypy/module/cpyext/test/test_<something>.py -k <testname>
```

This will build just enough of PyPy to run the test, which should take ~45 seconds the first time and less after that, and then run the test.

To check that the test is valid, you can do
```
python2 pytest.py -A --python=cpython3.11 pypy/module/cpyext/test/test_<something>.py -k <testname>
```

which will compile the test snippet into a c-extension module and run the test.

mattip:
Thanks @da-woods!

PR #5169, opened 2025-01-09
Fix the type of sendfunc

da-woods:
This fails to compile on C++ because it's more strict than C about matching exact function types..

### DIFF
--- a/pypy/module/cpyext/parse/cpyext_object.h
+++ b/pypy/module/cpyext/parse/cpyext_object.h
@@ -156,7 +156,7 @@ typedef enum {
     PYGEN_NEXT = 1,
 } PySendResult;
 
-typedef int (*sendfunc)(PyObject *iter, PyObject *value, PyObject **result);
+typedef PySendResult (*sendfunc)(PyObject *iter, PyObject *value, PyObject **result);
 
 typedef struct {
     unaryfunc am_await;


mattip:
Thanks @da-woods

mattip:
Whoops, this did not target `py3.10`, rather one of the branches uses in the transition from mercurial to git that we keep around for no real good reason. I cherry-picked it into the correct active branch `py3.10`. Sorry for not noticing. I also merged py3.10 into py3.11.

Maybe we should delete those branches from github.

mattip:
We need to parse the files in `pypy/module/cpyext/parse` to interop with RPython. It seems our parser cannot handle a function return type of `enum`. How can I reproduce the compilation failure you are seeing?

mattip:
You can see this fails by running a translation or an cpyext backend test.

da-woods:
I think this code reproduces the issue

```
#include "Python.h"

PySendResult mysendfunc(PyObject *iter, PyObject *value, PyObject **result); /* proto */

void wont_compile_on_cpp() {
    sendfunc fptr = mysendfunc;
}
```

(compiled with c++). Obviously it doesn't do anything useful, but it'll give an error. (Typed on my phone though, so not tested)

I'm fairly sure these work correctly on PyPy in C (where it's more liberal about pointers of different types). So the problem is just with the definition that appears in Python.h, not with the implementation.

mattip:
Right, but I would claim the problem is actually with `PySendResult mysendfunc(...` which should not be declared like that. I think the CPython declaration is subtly broken, see #5170

PR #5170, opened 2025-01-10
Make return type of `sendfunc` compatible with CPython

mattip:
In CPython, the [declaration of `sendfunc`](https://github.com/python/cpython/blob/baf65715fc9002e43cd0e1010b8dba9b4c84d503/Include/cpython/object.h#L127) is
```
typedef PySendResult (*sendfunc)(PyObject *iter, PyObject *value, PyObject **result);
```

where `PySendResult` is an [`enum`](https://github.com/python/cpython/blob/baf65715fc9002e43cd0e1010b8dba9b4c84d503/Include/object.h#L683). In PyPy we use `int` since
- the `sizeof(enum)` is undefined in C
- more practically, our parser, needed to reflect the C typedef into RPython, does not accept an enum there (maybe to prevent exactly this problem?).

In PR #5169, @da-woods submitted this change. I merged it but then realized there is a problem here so I am recreating the PR. I think the PR should rejected and code that uses the result of `sendfunc` should cast the result instead. @vstinner should this be considered a problem with CPython? Are there other public API functions that use enum return results?

### DIFF
--- a/pypy/module/cpyext/parse/cpyext_object.h
+++ b/pypy/module/cpyext/parse/cpyext_object.h
@@ -156,7 +156,7 @@ typedef enum {
     PYGEN_NEXT = 1,
 } PySendResult;
 
-typedef int (*sendfunc)(PyObject *iter, PyObject *value, PyObject **result);
+typedef PySendResult (*sendfunc)(PyObject *iter, PyObject *value, PyObject **result);
 
 typedef struct {
     unaryfunc am_await;


da-woods:
> the `sizeof(enum)` is undefined in C

I'm not sure if this is practically a problem from a C API point of view (although it may be a problem for PyPy depending its own implementation). 
The sizeof `int` is also implementation defined but practically this is part of the platform ABI so compilers on the same platform should end up agreeing. I'd assume (but I'm not absolutely sure) that an enum would be the same.

da-woods:
See https://github.com/pypy/pypy/pull/5171 for an alternative approach

vstinner:
CPython is trying to move away from enums in the C API: https://github.com/capi-workgroup/problems/issues/35

For existing APIs, I'm not sure if we should change them or not.

cc @encukou

da-woods:
@vstinner 

In case you haven't read the linked PRs and discussion for this:

The problem here was that Cython declared its `sendfunc` with a result type of `PySendResult`. C++ is quite strict about compatible functions types, and refused to compile when you assigned a function returning `PySendResult` to a function pointer type returning `int`.

I think that means it's quite difficult to change existing definitions without breaking people's C++ code. I've hopefully got a workaround for PyPy, but it is a workaround and might well not be acceptable in CPython.

(And that's ignoring the question of whether you can do it without breaking the stable ABI anywhere)

encukou:
This is part of stable ABI, so I don't think CPython can change it.

> The sizeof `int` is also implementation defined but practically this is part of the platform ABI so compilers on the same platform should end up agreeing. I'd assume (but I'm not absolutely sure) that an `enum` would be the same.

It's roughly the same for enum, but with a complication: the size of any particular enum can depend on its contents.

AFAIK, “in practice on major platforms”, it'll be just be `int` for the small enums CPython uses. It would be great to not rely on that assumption though.

Do you have some compiled C code, perhaps in tests, that could assert `sizeof(PySendResult as defined by CPython) == sizeof(int)` (ideally to fail a single test)?
It would be nice CPython could guarantee this, but I don't see that happening soon.

> Are there other public API functions that use enum return results?

I didn't check, a good starting point would be `git grep enum Include ':!Include/internal' ':!Include/pyexpat.h'` in the CPython repo.

vstinner:
> This is part of stable ABI, so I don't think CPython can change it.

Are you sure that it's part of the stable ABI? It's declared in `Include/cpython/object.h` which is excluded from the limited C API. And `PyTypeObject` structure members are excluded from the stable ABI. Slots can only be retrieved by `PyType_GetSlot()`.

vstinner:
Ah, `PyIter_Send()` uses the enum as the return type and it's part of the stable ABI.

```c
PyAPI_FUNC(PySendResult) PyIter_Send(PyObject *, PyObject *, PyObject **);
```

encukou:
Yes, and you can also set `Py_am_send` to a function with the correct signature (which is [documented](https://docs.python.org/3/c-api/typeobj.html#c.PyAsyncMethods.am_send)). C code doesn't normally need to use the *name* `sendfunc` at all.

mattip:
#5171 was merged, and it seems the declaration cannot be changed in CPython before ABI4 or so. Closing, thanks for the clarifications and fixes.

PR #5171, opened 2025-01-11
Fix sendfunc definition second go

da-woods:
Alternative approach to sendfunc. Hopefully supersedes #5170.

I do plan to write a test for this fairly soon, but proposing this PR now just to see if it works at all. So treat it as draft until the test.

### DIFF
--- a/pypy/module/cpyext/parse/cpyext_object.h
+++ b/pypy/module/cpyext/parse/cpyext_object.h
@@ -152,10 +152,10 @@ typedef struct {
 /* Result of calling PyIter_Send
 
 Note: small deviation from CPython definitions here:
-PyPy wants to return and int from "sendfunc" because it knows sizeof(int)
+PyPy wants to return an int from "sendfunc" because it knows sizeof(int)
 and doesn't know sizeof(enum PySendResult).
 Therefore introduce the symbols like PYGEN_RETURN into the global namespace
-with an anoymous enum, but define PySendResult as int so that funnctions
+with an anoymous enum, but define PySendResult as int so that functions
 that are defined as returning PySendResult will be compatible with the definition
 of sendfunc (especially on C++).
 The only thing this breaks is C++ code that writes PySendResult::PYGEN_RETURN.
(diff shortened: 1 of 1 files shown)


pypy/module/cpyext/parse/cpyext_object.h
```diff
@@ -149,12 +149,24 @@ typedef struct {
...
+Note: small deviation from CPython definitions here:
+PyPy wants to return and int from "sendfunc" because it knows sizeof(int)
+and doesn't know sizeof(enum PySendResult).
+Therefore introduce the symbols like PYGEN_RETURN into the global namespace
+with an anoymous enum, but define PySendResult as int so that funnctions
+that are defined as returning PySendResult will be compatible with the definition
+of sendfunc (especially on C++).
+The only thing this breaks is C++ code that writes PySendResult::PYGEN_RETURN.
+*/
 typedef enum {
```
da-woods, line 163:
```suggestion
enum {
```

pypy/module/cpyext/parse/cpyext_object.h
```diff
@@ -149,12 +149,24 @@ typedef struct {
...
 
-/* Result of calling PyIter_Send */
-typedef enum {
+/* Result of calling PyIter_Send
+
+Note: small deviation from CPython definitions here:
+PyPy wants to return and int from "sendfunc" because it knows sizeof(int)
+and doesn't know sizeof(enum PySendResult).
+Therefore introduce the symbols like PYGEN_RETURN into the global namespace
+with an anoymous enum, but define PySendResult as int so that funnctions
```
mattip, line 158:
```suggestion
PyPy wants to return an int from "sendfunc" because it knows sizeof(int)
and doesn't know sizeof(enum PySendResult).
Therefore introduce the symbols like PYGEN_RETURN into the global namespace
with an anoymous enum, but define PySendResult as int so that functions
```

mattip (approved):
Seems to work. I wonder if anyone is using `PySendResult::PYGEN_RETURN` and friends.

da-woods:
Unfortunately I don't think I can write a good test because:
* `PyIter_Send` isn't implemented in PyPy so there's not a lot you can do with the `am_send` member except initialize it.
* I can't get compilation to break in C mode (only in C++ mode) and none of the existing tests seem to use C++ so I'd need to add a reasonable amount of infrastructure to get something to build.

So I think this is probably as good as I can do for now.

mattip:
Thanks @da-woods, I think this is as good as we can get.

Issue #5177, opened 2025-01-21
native symbol names in vmprof, when running nightly builds


cfbolz:
originally @Cskorpion and I thought that #5114 would fix the problem of showing symbols of native functions in vmprof. However, it turns out it only works for locally built pypys, the nightly builds still have the same problem.

To fix this we would have to ship a recent version of libunwind. @Cskorpion has prepared the necessary changes to the release script so that we ship the `.so`, and will open a PR.  he has also changed the dockerfile of the buildbot container to build the correct libunwind, I'll open a PR for that.

cfbolz:
@mattip should we do the build of a recent libunwind in the Dockerfile, or rather in ``lib_pypy/pypy_tools/build_cffi_imports.py``?

mattip:
The advantage of the docker build is that we only do it once. But if the time to build is small, maybe it makes sense to remain flexible and rebuild on every buildbot run. The docker files are here https://foss.heptapod.net/pypy/buildbot/-/tree/branch/default/docker?ref_type=heads

cfbolz:
I think the docker files also make a bit more sense because `build_cffi_imports.py` is really meant for things written with cffi, which is no the case for `_vmprof`. I'll give a PR to the Dockerfiles a go then.

cfbolz:
here's my pr for the dockerfiles: https://foss.heptapod.net/pypy/buildbot/-/merge_requests/7/diffs

mattip:
The PR for the dockerfiles was merged, and the images are in use on x86_64, x86_32, and aarch64. Although it seems `VMP_SUPPORTS_NATIVE_PROFILING` will not be defined for aarch64. This comes from upstream? https://github.com/pypy/pypy/blob/eb65c6bab13b77c8cbcbb48b478aa467b8ba4243/rpython/rlib/rvmprof/src/shared/vmprof.h#L51-L55

mattip:
#5180 was merged, and the libunwind from the buildbots is packaged with the portable builds on linux. Latest nightly builds should be available in a couple of hours. Can you check that all the pieces are in place to show native symbols?

Cskorpion:
As for the nightly build `2025-01-24` (x86_64 linux) `libunwind.so` is included in the lib folder and is loaded at runtime to resolve names.

Example profile in the Firefox-Profiler UI:

![Image](https://github.com/user-attachments/assets/81472e7b-a02c-49d0-b241-347ca9d077bd)

cfbolz:
[...]

PR #5180, opened 2025-01-22
VMProf: find native symbol names via libunwind & ship libunwind 

Cskorpion:
This will allow VMProf to find native function names via libunwind's function `unw_get_proc_name_by_ip`, which usually finds names, that `dladdr` and `libbacktrace` won't find.

The changes in `make_portable.py` will add `libunwind.so` to the list of dependencies, so that it will be shipped with the PyPy shared object and can be loaded at runtime in `symboltable.c`.

As described by @cfbolz in #5177, a buildbot would need to build a recent version of libunwind for this to work.

### DIFF
--- a/pypy/tool/release/make_portable.py
+++ b/pypy/tool/release/make_portable.py
@@ -122,9 +122,10 @@ def make_portable():
 
     deps = gather_deps(binaries)
 
-    # Add libunwind manually, because its no binarys dependency (we load libunwind at runtime with dlsym)
-    # Must be set to the path where libunwind is installed to in dockerfile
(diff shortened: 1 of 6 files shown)


mattip:
LGTM thanks. Should we backport some of this to upstream vmprof? Is it useful for CPython as well? If not, we need to adjust the [test for shared files](https://github.com/pypy/pypy/blob/eb65c6bab13b77c8cbcbb48b478aa467b8ba4243/rpython/rlib/rvmprof/test/test_file.py#L15) and add the new files to the list of excluded ones.

mattip:
If there are more commits needed, please start a new branch from `'main'` rather than continuing from `'vmprof-native-profiling-name'`

Issue #5186, opened 2025-01-28
libunwind crashes when running vmprof with native profiling


cfbolz:
unfortunately the new code introduced in #5180 leads to segfaults when using vmprof with native profiling. I think we have to fix this before the release. we have a few option:
- simply revert the feature
- try to understand/fix the libunwind bug (@Cskorpion is looking into this a bit)
- stop using libunwind for symbolication and do it ourselves instead (we have a lot of the infrastructure in place, at least on linux)

mattip:
Is there a minimum reproducer?

cfbolz:
we're working on it. it happens with some percent probability when I use vmprof to profile rpython:

`pypy --jit off -m vmprof ../../rpython/bin/rpython --source ../../rpython/translator/goal/targetrpystonedalone.py`

cfbolz:
@mattip I'm confused by https://github.com/pypy/pypy/commit/bb7da9fdf62992f267ff69323b3b20a33fce375f#diff-b563068d6297a434c60cfb68af95e350b788ff3b4070702d42588859d274f896, it looks like you removed the feature on main in that commit? (that might be what we need to do anyway now, of course).

cfbolz:
@Cskorpion managed to reproduce it:

```c
#include <link.h>
#include <stdlib.h>
#include <stdio.h>
//#include "libunwind.h"

static void* unw_local_addr_space = NULL;
static int (*unw_get_proc_name_by_ip)(void *, u_int64_t, char *, size_t, u_int64_t *, void *) = NULL;

int main(void) {
    // This code should reproduce the segfault occuring when using vmprof to profile code on pypy mentioned in https://github.com/pypy/pypy/issues/5186
    // The segfault occurs when trying to resolve the name of a function, that is not exposed in the symboltable of its correspoonding shared object
    // And (so far) only for libffi functions 

    // Load libunwind, unw_get_proc_name_by_ip and unw_local_addr_space 
    // A self built, recent version of libunwind is required
    void * handle_libunwind = NULL;
    if ((handle_libunwind = dlopen("libunwind.so", RTLD_NOW)) == NULL) {
        printf("couldn't open libunwind \n");
        return 1;
    }
    unw_get_proc_name_by_ip = dlsym(handle_libunwind, "_ULx86_64_get_proc_name_by_ip");
    unw_local_addr_space = dlsym(handle_libunwind, "_ULx86_64_local_addr_space");
[...]

PR #5190, opened 2025-01-30
Vmprof manual symbolify

cfbolz:
#5186: instead of using libunwind to symbolify (which crashes), parse the elf files ourselves (which we do for the remote debugger already anyway). This probably has some not implemented corner cases, but at least it can't segfault.


mattip:
This replaces #5180 right?

cfbolz:
> This replaces #5180 right?

yep. as a next step we can remove some of that code in `symboltable.c` again. it's not being used in `_vmprof` after this PR, but I wanted to land this one before removing the other code, in case the nightly builds show a problem.

PR #5192, opened 2025-02-04
revert rvmprof resolve_address via libunwind

mattip:
Revert rvmprof changes in #5180 to create a rvmprof `resolve_address` function that uses `libunwind`. Related to #5186


mattip:
Lastest nightly run with the `/modules/_vprof` ELF-parsing code passed cleanly, so this can go in.

PR #5200, opened 2025-02-08
Adjust sqlite_*int64 types to match upstream

mgorny:
The sqlite3 int64 types are defined as:

```
  typedef long long int sqlite_int64;
  typedef unsigned long long int sqlite_uint64;
```

This technically is distinct from `int64_t` and `uint64_t` (that are defined as plain `[unsigned] long int` on 64-bit architectures). Adjust the definitions in `_sqlite3_build.py` to match.

### DIFF
--- a/lib_pypy/_sqlite3_build.py
+++ b/lib_pypy/_sqlite3_build.py
@@ -168,8 +168,8 @@ typedef ... sqlite3;
 typedef ... sqlite3_stmt;
 typedef ... sqlite3_context;
 typedef ... sqlite3_value;
-typedef int64_t sqlite3_int64;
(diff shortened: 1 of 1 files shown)


cfbolz:
the change makes sense, but it should target the py3.10 branch. The same problem is there. I'll do it manually by cherry-picking the change.

cfbolz:
done: https://github.com/pypy/pypy/commit/9031a0d591b40b91e975c547279cb09f8915e837
(and merged to py3.11 too)

mgorny:
Thanks! Curious enough, I didn't experience the compile failure with PyPy3.10.

Issue #5199, opened 2025-02-07
`-Wlto-type-mismatch` with pypysig_counter in 7.3.18


thesamesam:
When building pypy3_10-7.3.18 in Gentoo with LTO (and using `-Werror=lto-type-mismatch`), I got:
```
/var/tmp/portage/dev-python/pypy3_10-exe-7.3.18/work/pypy3.10-v7.3.18-src/rpython/translator/c/src/signals.h:46:41: error: type of ‘pypysig_counter’ does not match original declaration [-Werror=lto-type-mismatch]
   46 | RPY_EXPORTED struct pypysig_long_struct pypysig_counter;
      |                                         ^
signals.c:45:28: note: ‘pypysig_counter’ was previously declared here
   45 | struct pypysig_long_struct pypysig_counter = {
      |                            ^
lto1: some warnings being treated as errors
lto-wrapper: fatal error: x86_64-pc-linux-gnu-gcc returned 1 exit status
compilation terminated.
/usr/lib/gcc/x86_64-pc-linux-gnu/15/../../../../x86_64-pc-linux-gnu/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make: *** [Makefile:1106: pypy3.10-c] Error 1
```

7.3.17 was fine. Full log (xz-compressed): https://bugs.gentoo.org/attachment.cgi?id=918312.

thesamesam:
Thanks, let me have a look.

thesamesam:
It's similar to https://gcc.gnu.org/PR114218. The issue is differing values for `PATH_MAX`. If you include `limits.h` before including `signals.h`, it's fine, but if you don't, you get the fallback value of 1024.

This fixes it, but it's likely not portable enough:
```
--- a/rpython/translator/c/src/signals.h
+++ b/rpython/translator/c/src/signals.h
@@ -1,6 +1,7 @@
 #ifndef _PYPY_SIGNALS_H
 #define _PYPY_SIGNALS_H

+#include <limits.h>
 #include "src/precommondefs.h"
```

cfbolz:
ah, I see! thank you. given that signals.c unconditionally includes limits.h, we should be fine having it in signals.h (probably *after* precommondefs.h). does that fix things for you?

thesamesam:
That works. Submitted https://github.com/pypy/pypy/pull/5202.

PR #5202, opened 2025-02-10
Fix LTO issue with pypysig_counter

thesamesam:
If `limits.h` isn't included, we get two values in use for `PATH_MAX` -
one is the fallback defined by `signals.h`, and the other is the one
that `signals.c` sees when it includes `limits.h`. This is a problem
with LTO if another object included limits.h and picked up a different
definition of `pypysig_counter`, showing up as -Wlto-type-mismatch.

Fix that by including `limits.h` unconditionally. signals.c already
does that so it should be fine.

Closes: #5199

### DIFF
--- a/rpython/translator/c/src/signals.h
+++ b/rpython/translator/c/src/signals.h
@@ -24,10 +24,6 @@ int pypysig_poll(void);   /* => signum or -1 */
 RPY_EXTERN
 void pypysig_pushback(int signum);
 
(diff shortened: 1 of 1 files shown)


thesamesam:
I think the CI failure is unrelated (it seems flaky looking at recent commits).

thesamesam:
@cfbolz can you take a look?

thesamesam:
Thank you!

PR #5235, opened 2025-03-02
py3.11: raise `SyntaxError` on null bytes, to match CPython 3.11

### LINKED ISSUE #5234: PyPy3.11: "source code string cannot contain null bytes" exception should be `SyntaxError` in Python 3.11.4+
Not a big deal, but should be trivial to fix. I can make a patch later, since I can't test it right now.

Relevant code from pycodestyle's test suite:

```py
        if sys.version_info < (3, 11, 4):  # pragma: <3.11 cover
            expected = ["stdin:1:1: E901 ValueError: source code string cannot contain null bytes"]  # noqa: E501
        elif sys.version_info < (3, 12):  # pragma: <3.12 cover  # pragma: >=3.11 cover  # noqa: E501
            expected = ["stdin:1:1: E901 SyntaxError: source code string cannot contain null bytes"]  # noqa: E501
        else:  # pragma: >=3.12 cover
            expected = [
                "stdin:1:1: E901 SyntaxError: source code string cannot contain null bytes",   # noqa: E501
                "stdin:1:1: E901 TokenError: source code cannot contain null bytes",   # noqa: E501
            ]
```

mgorny:
Raise `SyntaxError` rather than `ValueError` on null bytes in source code, to match the behavior of Python 3.11.4+.

Fixes #5234

### DIFF
--- a/lib-python/3/test/test_ast.py
+++ b/lib-python/3/test/test_ast.py
@@ -864,8 +864,7 @@ class AST_Tests(unittest.TestCase):
         check_limit("a", "*a")
 
     def test_null_bytes(self):
-        # PyPy raises a ValueError
-        with self.assertRaises((SyntaxError, ValueError),
+        with self.assertRaises(SyntaxError,
             msg="source code string cannot contain null bytes"):
             ast.parse("a\0b")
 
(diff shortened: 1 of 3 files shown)


mattip:
Thanks

mgorny:
Thank you!

PR #5241, opened 2025-03-07
Fix `datetime.*.fromisoformat()` to reject spaces in fraction part

mgorny:
Fix pure Python implementation of `fromisoformat()` to reject spaces in fractional part of time specifications.  This matches the behavior of the C implementation in CPython, and prevents the method from parsing them incorrectly.

This fixes Django's test suite by enabling a fallback to their own (working) parser.

Backports: https://github.com/python/cpython/pull/130962


mgorny:
Sending it now, since I've prepared the backport for Gentoo already — but we may want to wait for CPython review before merging.

mattip:
Thanks

Issue #5229, opened 2025-02-26
Translation broken on 32-bit ARM on older platforms: assert OFF_T_SIZE == rffi.sizeof(rffi.LONGLONG)


stefanor:
Translation of 7.3.19 (3.11) on Ubuntu 20.04 and 22.04 on 32-bit ARM (armhf) fail with:

```
[translation:info] 2.7.18 (default, Feb 26 2025, 21:30:15) 
[GCC 11.4.0]
[platform:msg] Set platform with 'host' cc=None, using cc='gcc', version='Unknown'
[translation:info] Translating target as defined by targetpypystandalone
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/rpython/bin/rpython", line 20, in <module>
    main()
  File "/<<PKGBUILDDIR>>/rpython/translator/goal/translate.py", line 219, in main
    targetspec_dic, translateconfig, config, args = parse_options_and_load_target()
  File "/<<PKGBUILDDIR>>/rpython/translator/goal/translate.py", line 155, in parse_options_and_load_target
    targetspec_dic = load_target(targetspec)
  File "/<<PKGBUILDDIR>>/rpython/translator/goal/translate.py", line 97, in load_target
    mod = __import__(specname)
  File "targetpypystandalone.py", line 7, in <module>
    from pypy.interpreter import gateway
  File "/<<PKGBUILDDIR>>/pypy/interpreter/gateway.py", line 19, in <module>
    from pypy.interpreter.eval import Code
  File "/<<PKGBUILDDIR>>/pypy/interpreter/eval.py", line 6, in <module>
    from pypy.interpreter.baseobjspace import W_Root
  File "/<<PKGBUILDDIR>>/pypy/interpreter/baseobjspace.py", line 13, in <module>
    from rpython.rlib.buffer import StringBuffer
  File "/<<PKGBUILDDIR>>/rpython/rlib/buffer.py", line 16, in <module>
    from rpython.rlib import rawstorage
  File "/<<PKGBUILDDIR>>/rpython/rlib/rawstorage.py", line 49, in <module>
    cpuname = detect_cpu.autodetect()
  File "/<<PKGBUILDDIR>>/rpython/jit/backend/detect_cpu.py", line 117, in autodetect
    return detect_model_from_host_platform()
  File "/<<PKGBUILDDIR>>/rpython/jit/backend/detect_cpu.py", line 105, in detect_model_from_host_platform
[...]

PR #5254, opened 2025-03-24
Get 32-bit ARM translating on old platforms again

stefanor:
This applies to the 3.10 and 3.11 branches.

Fixes: #5229


mattip:
Merged to py3.11 in 66b5c42b46f

PR #5260, opened 2025-04-10
perf: optimize splitlines method performance

MoYeRanqianzhi:
Improve performance of splitlines method for both regular strings and Unicode strings:

- Add fast path for empty strings
- Pre-calculate maximum possible line count and pre-allocate result list
- Reduce number of list resizing operations
- Optimize memory usage and allocation patterns
- Improve main loop structure and character comparison logic
- Add clear comments explaining the implementation

This addresses the splitlines performance issue mentioned in the TODO file, with significant improvements for large multi-line strings.

Modified files:
- rpython/rtyper/rstr.py
- pypy/objspace/std/unicodeobject.py

Thank you for contributing to PyPy. Please be sure to target one of our active branches:
- `main` for pypy2.7 and rpython changes
- `py3.10` for the legacy python3.10 branch
- `py3.11` for the current development branch targeting python3.11

Do not target the `branch/*` branches, they are artifacts of our migration from mercurial to git.

### DIFF
--- a/pypy/objspace/std/unicodeobject.py
+++ b/pypy/objspace/std/unicodeobject.py
@@ -664,19 +664,39 @@ class W_UnicodeObject(W_Root):
     def descr_splitlines(self, space, keepends=False):
         value = self._utf8
         length = len(value)
-        strs_w = []
+
+        # Fast path for empty strings
+        if length == 0:
+            return space.newlist([])
+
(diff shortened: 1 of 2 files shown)


mattip:
What kind of speed up do you get with these changes?

cfbolz:
the approach looks reasonable! I agree with @mattip though, could you please tell us what measurements you did and the improvements you saw?

PR #5262, opened 2025-04-21
Remove some duplicate code from `_cffi_ssl` module

mgorny:
Remove some duplicate code that I've discovered in `_cffi_ssl` module while debugging https://github.com/python-trio/trio/issues/3253.


mattip:
Thanks @mgorny

mgorny:
Thanks!

PR #5263, opened 2025-04-21
[_cffi_ssl] Add missing LIB_ERR_SYS handling

mgorny:
Add missing handlers for `LIB_ERR_SYS` in `pyssl_error()`.  This was added to CPython in python/cpython@f4b31edf2d9d72878dab1f66a36913b5bcc848ec, and seems to be required for OpenSSL 3.4+.

Adding this handler fixes flaky test failures reported in python-trio/trio#3253.


mattip:
Thanks @mgorny

PR #5280, opened 2025-05-23
Fix EINTR handling in time.sleep() on Mac

rlamy:
nanosleep() is documented as returning -1 on error and setting errno.


mattip:
Thanks @rlamy

PR #5281, opened 2025-05-28
Fix segfault when passing invalid closure to exec()

rlamy:


mattip:
Thanks. Where did that come up?

rlamy:
That was causing a segfault in `lib-python/3/test/test_builtin` - which was hiding a bunch of errors, it turns out.

PR #5282, opened 2025-05-29
Translate variant with --revdb in “Translate PyPy” GitHub workflow.

manueljacob:
Before this change, translating RevDB was not tested in CI, leading to occasional breaks. Also, having the translated RevDB-enabled PyPy variant available as an Artifact makes it easier for people to try it.

### DIFF
--- a/.github/workflows/translate.yml
+++ b/.github/workflows/translate.yml
@@ -13,6 +13,7 @@ permissions:
 jobs:
   build:
 
+    name: build on ${{ matrix.os }}${{ matrix.revdb && ' (with RevDB)' || '' }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false


cfbolz:
Makes sense. Are there any app level revdb tests that we should run on the translated binary? Do we need to hold off merging until we find out what the 3.11 problem is you mentioned on irc?

manueljacob:
> Makes sense. Are there any app level revdb tests that we should run on the translated binary?

In a quick look, I found only `AppTestBufferedReaderOnRevDB`. If I recall correctly, that would also run untranslated. I’m not sure whether doing the analog to `pypy-c-jit-linux-x86-64` (running all app-level and lib-python tests on the translated binary) would be a good idea in the moment. Many tests will fail because they are not supported by RevDB. Independent from RevDB, will this testing infrastructure stay on buildbot long-term?

> Do we need to hold off merging until we find out what the 3.11 problem is you mentioned on irc?

Considering that merging this branch into the py3.11 branch will introduce a failing job in the moment, deferring the merge seems like a good idea.

manueljacob:
When disabling the `_hpy_universal` module, RevDB now works on the py3.11 branch.

I’ve prepared a branch targetting py3.11 containing these CI changes and disabling the `_hpy_universal` module in CI on py3.11: https://github.com/pypy/pypy/compare/py3.11...py3.11-ci-revdb

If https://github.com/pypy/pypy/actions/runs/15364349940 succeeds, I think that both branches can be merged.

PR #5285, opened 2025-05-30
Fix test failures in test_builtin.py

rlamy:
This implements these 2 CPython PRs:
* https://github.com/python/cpython/pull/25863
* https://github.com/python/cpython/pull/15592

and fixes the test file for expected differences with CPython (some which were handled before but got lost in the last stdlib update).

### DIFF
--- a/lib-python/3/test/test_builtin.py
+++ b/lib-python/3/test/test_builtin.py
@@ -2035,7 +2035,11 @@ class BuiltinTest(unittest.TestCase):
             array.clear()
             yield b'A'
             yield b'B'
-        self.assertRaises(BufferError, array.join, iterator())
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @rlamy

mattip:
This changed the error in `zip(*args)` from `TypeError: zip argument #1 must support iteration` to the CPython compatible `TypeError: 'NoneType' object is not iterable` (where args is `[None, (), (), (), ()]`). I am not sure the CPython-compatible error message is better.

mattip:
Ahh, nevermind. The error message was hashed over, [pronounced better](https://bugs.python.org/issue37976#msg350848), and intentionally changed to what was done here.

PR #5287, opened 2025-05-31
added from __future__ import print_function for better python3 compatibility

auvipy:


pypy/bin/checkmodule.py
```diff
@@ -14,21 +15,21 @@
 
 def main(argv):
     if len(argv) != 2:
-        print >> sys.stderr, __doc__
+        print( >> sys.stderr, __doc__)
```
mattip, line 18: This runs? It looks really wrong to me. I think it (and the others) should be
```suggestion
        print(__doc__, file=sys.stderr)
```

auvipy:
I was actually trying them manually. thanks for the suggestions! I will modify them come back to you soon

pypy/bin/checkmodule.py
```diff
@@ -14,21 +15,21 @@
...
-        print >> sys.stderr, __doc__
+        print( >> sys.stderr, "Bad command line")
+        print(__doc__, file=sys.stderr)
         sys.exit(1)
     if os.path.sep in modname:
         if os.path.basename(modname) == '':
             modname = os.path.dirname(modname)
         if os.path.basename(os.path.dirname(modname)) != 'module':
-            print >> sys.stderr, "Must give '../module/xxx', or just 'xxx'."
+            print( >> sys.stderr, "Must give '../module/xxx', or just 'xxx'.")
```
mattip, line 32: What about this one? And there are more in the other files...

PR #5289, opened 2025-06-02
feat: add missing constants in zlib

RFCreate:
I have added some missing constants for the `zlib` module that are present in the CPython module. See [CPython commit](https://github.com/python/cpython/commit/bc3f2289b9007396bfb7f986bee477b6176c1822).

I am not sure what branch to target, this applies to `main`, `py3.10`, and `py3.11`; if so I can open other PRs for the other branches or what it is needed.

### DIFF
--- a/pypy/module/zlib/moduledef.py
+++ b/pypy/module/zlib/moduledef.py
@@ -39,10 +39,27 @@ objects support decompress() and flush()."""
 
 
 for _name in """
-    MAX_WBITS  DEFLATED  DEF_MEM_LEVEL
-    Z_BEST_SPEED  Z_BEST_COMPRESSION  Z_DEFAULT_COMPRESSION
-    Z_FILTERED  Z_HUFFMAN_ONLY  Z_DEFAULT_STRATEGY
-    Z_FINISH  Z_NO_FLUSH  Z_SYNC_FLUSH  Z_FULL_FLUSH
+    MAX_WBITS
+    DEFLATED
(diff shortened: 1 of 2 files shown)


RFCreate:
If I understand correctly, this needs to extend the `constantnames` with the missing constants which might be confusing or rewrite them all,  both options in `interp_zlib.py`.

Then do all this code from rpython again in `interp_zlib.py`.
```
for _name in constantnamesNEW:
    setattr(rzlib.SimpleCConfig, _name, rzlib.rffi_platform.ConstantInteger(_name))

config = rzlib.rffi_platform.configure(rzlib.SimpleCConfig)
for _name in constantnamesNEW:
    globals()[_name] = config[_name]
```

Finally in `moduledef.py` do something like this:
```
interpleveldefs = {
    'CONST1': 'interp_zlib.CONST1',
    'CONST2': 'interp_zlib.CONST2',
    'CONST3': 'interp_zlib.CONST3',
    'CONST4': 'interp_zlib.CONST4',
    ...
}
```

Its duplicating code and making hard to read IMO.

mattip:
It seems these are the missing constants (by comparing cpython3.11 to pypy3.11).

`'DEF_BUF_SIZE', 'ZLIB_RUNTIME_VERSION', 'ZLIB_VERSION', 'Z_BLOCK', 'Z_FIXED', 'Z_NO_COMPRESSION', 'Z_PARTIAL_FLUSH', 'Z_RLE', 'Z_TREES'`

PR #5292, opened 2025-06-05
Add missing audit event for `sys._getframe()`

rlamy:
CPython test added in https://github.com/python/cpython/pull/94928

Also, fix `@pytest.mark.pypy_only` to work in `-D` apptests with non-ancient pytest versions.

### DIFF
--- a/pypy/conftest.py
+++ b/pypy/conftest.py
@@ -44,6 +44,13 @@ except AttributeError:
     # no need to dedent on py3.10+
     pass
 
+def get_marker(item, name):
+    try:
+        return item.get_closest_marker(name=name)
+    except AttributeError:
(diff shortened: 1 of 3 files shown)


mattip:
On the buildbots [an untranslated settrace](https://buildbot.pypy.org/summary/longrepr?testname=AppTestSysSettracePortedFromCpython.%28%29.test_sys_settrace&builder=own-linux-x86-64&build=10576&mod=pypy.module.sys.test.test_sysmodule) test failed, but it passes locally.

rlamy:
I can reproduce it with `py.test pypy/module/sys`. I think there's some interaction between audit and tracing tests.

PR #5301, opened 2025-06-29
Replace a test class that has only a single test method with function

Neppord:
This is a warmup for me trying to improve the parsing library in rlib


Neppord:
What is this missing for it to be merged?

Is it to large?

cfbolz:
sorry, needed to find some time to remember how rlib/parsing works :grimacing:

PR #5305, opened 2025-07-03
remove named-branches action

mattip:
I think the named branches integration is no longer needed. We put it in so people who wanted to use mercurial could still do so. @Julian, @arigo is it OK to remove this?


arigo:
@mattip I guess that is some hack that was done during the transition from mercurial to keep some workflow possible?  No, I never used it, feel free to remove as far as I'm concerned.

PR #5307, opened 2025-07-05
Remove `pycparser` dep from `cffi` dist-info

### LINKED ISSUE #5306: PyPy3.11: newly created venv has unsatisfied dependencies
```
$ pypy3.11 -m venv .venv
$ .venv/bin/pip check
cffi 1.18.0.dev0 requires pycparser, which is not installed.
```

I'm guessing this is due to https://github.com/python-cffi/cffi/commit/22d4fd8f70bcd7a62565222c5c578d84d746924e which made `pycparser` dependency unconditional.

mgorny:
Remove the incorrect `pycparser` dependency from `cffi` dist-info. PyPy is vendoring `pycparser` inside `cffi`, and not as a global package.  Having an explicit dependency meant that `pip check` would fail in environments without `pycparser` manually installed, in particular in a fresh venv.

Fixes #5306


mgorny:
Thanks!

PR #5331, opened 2025-09-24
Jit backend longevity without dict

cfbolz:
make the JIT backend quite a bit faster (roughly 50% in my measurements) by using a lot fewer dictionaries.

It's possible that this is causing issues for some of the backends we don't have a buildbot for (s390x/riscv), even though I tried to fix them.


mattip:
On x86_64 rlib/unicodedata/test/test_dawg.py is newly failing, it timed out.

cfbolz:
It has always timed out occasionally 😕. Maybe now is a good time to finally investigate it

PR #5364, opened 2026-01-08
Add sys._getframemodulename() implementation.

manueljacob:


mattip:
Please add py3.12 to the actions workflow https://github.com/pypy/pypy/blob/376cb12babe6bc8f43a550d2981a322513be24de/.github/workflows/translate.yml#L5 and also the rpython workflow so that CI will run on the branch

manueljacob:
Good point! We could generalize it to "py3.[0-9][0-9]" or even "py3.[0-9]+" (the syntax is defined [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet)). What do you think?

mattip:
Sure

manueljacob:
Done in #5365. If you’re fine with the changes, I can do the necessary merges / rebases to put this branch here on top of the CI changes.

PR #5372, opened 2026-02-04
Add missing `_varname_from_oparg` method to code objects

BarrensZeppelin:
The fallback was insufficient for disassembling some of the code I'm generating on the py3.12 branch(es).


mattip:
Some untranslated tests in would be nice in `pypy/interpreter/test/test_code.py`. See the examples there how to `exec` some source code and check its attributes and methods.

mattip (commented):
LGTM. @cfbolz any thoughts?

PR #5378, opened 2026-02-26
fix: replace 730 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


mattip:
No thanks. At this point all this can be considered legacy or vendored code and we will not be refactoring it.

PR #5384, opened 2026-03-10
Fix object.__format__ to check format spec before string conversion

### LINKED ISSUE #5383: `object.__format__` raises wrong exception with non-empty format spec
`object.__format__` behaves incompatibly with CPython when a non-empty format spec is given.
Found via differential fuzzing between CPython and PyPy.

## Environment
- PyPy 7.3.20 (Python 3.11.13)
- CPython 3.11.14
- OS: Debian 12

## Reproduction

```python
class A:
    def __repr__(self):
        raise ValueError

format(A(), "abc")
```

### Output

**PyPy**:
```
ValueError
```

**CPython**:
```
TypeError: unsupported format string passed to A.__format__
```

## Root Cause

In PyPy's [`descr___format__`](https://github.com/pypy/pypy/blob/6a2789b8328bfe9d4261681280bb7c35839ec3b9/pypy/objspace/std/objectobject.py#L285), the object is converted to a string before checking if the format spec is non-empty.
CPython's [`object___format___impl`](https://github.com/python/cpython/blob/cd1c3a6342869b7346c1b5c27b8de9c6ef9c4e69/Objects/typeobject.c#L5491) checks the format spec length first and never calls `__repr__`.

jseop-lim:
## Summary
- Check format spec length before converting object to string in `object.__format__`, matching CPython behavior.

## Details

Fixes #5383

- Moved the `format_spec` non-empty check before the `__str__`/`__repr__` conversion in `descr___format__`. Previously, the conversion happened first, so errors from `__repr__`/`__str__` would propagate instead of the expected `TypeError`.
- Added test case: `format()` with non-empty spec on an object whose `__str__` raises `ValueError` → should raise `TypeError`.

### DIFF
--- a/pypy/module/__builtin__/test/test_format.py
+++ b/pypy/module/__builtin__/test/test_format.py
@@ -23,3 +23,9 @@ class AppTestFormat(object):
                     raises(TypeError, format, cls(), fmt_str)
                 else:
                     format(cls(), fmt_str)  # does not raise
+
+        class D:
+            def __str__(self):
+                raise ValueError
+
+        raises(TypeError, format, D(), 's')
--- a/pypy/objspace/std/objectobject.py
+++ b/pypy/objspace/std/objectobject.py
@@ -284,15 +284,19 @@ def descr__reduce_ex__(space, w_obj, proto):
 
 def descr___format__(space, w_obj, w_format_spec):
     if space.isinstance_w(w_format_spec, space.w_unicode):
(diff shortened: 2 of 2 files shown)


jseop-lim:
Thanks @mattip for merging! It's an honor to make my first contribution to PyPy.

No, I found this through fuzzing. Details on how it was discovered are in #5383.
Glad it got fixed!

PR #5416, opened 2026-04-09
gh-5344: Fix `raise` ignoring user modifications to `__traceback__`

### LINKED ISSUE #5344: Setting `Exception.__traceback__` to `None` does not disable traceback
The problem can be reproduced with this code:

```python
try:
    raise ValueError("err")
except ValueError as err:
    err.__traceback__ = None
    raise
```

When the above is run with CPython 3.11 or newer, you only get the error message and no traceback:
```python
ValueError: err
```
With PyPy 3.11 you get the traceback as well:
```python
  File "...", line 2, in <module>
    raise ValueError("err")
ValueError: err
```

StanFromIreland:
Now we get:
```pypy
>>>> try:
....     raise ValueError('err')
.... except ValueError as err:
....     err.__traceback__ = None
....     raise
.... 
ValueError: err
```

Closes #5344.
CPython 2.7 also has this bug, so this PR targets py3.11 for consistency.

### DIFF
--- a/pypy/interpreter/pyopcode.py
+++ b/pypy/interpreter/pyopcode.py
@@ -705,7 +705,7 @@ class __extend__(pyframe.PyFrame):
             w_value = last_operr._w_value
             if w_value is not None and isinstance(w_value, W_BaseException):
                 w_tb = w_value.w_traceback
-                if w_tb is None or space.is_w(w_tb, space.w_None):
+                if space.is_none(w_tb):
                     last_operr.set_traceback(None)
                 else:
                     from pypy.interpreter.pytraceback import PyTraceback
--- a/pypy/interpreter/pyopcode.py
+++ b/pypy/interpreter/pyopcode.py
@@ -699,6 +699,18 @@ class __extend__(pyframe.PyFrame):
             if last_operr is None:
                 raise oefmt(space.w_RuntimeError,
                             "No active exception to reraise")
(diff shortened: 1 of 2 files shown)


pypy/interpreter/pyopcode.py
```diff
@@ -699,6 +699,18 @@ def RAISE_VARARGS(self, nbargs, next_instr):
             if last_operr is None:
                 raise oefmt(space.w_RuntimeError,
                             "No active exception to reraise")
+            # sync the exceptions __traceback__ back to the
+            # OperationError, in case user code modified it
+            from pypy.module.exceptions.interp_exceptions import W_BaseException
+            w_value = last_operr._w_value
+            if w_value is not None and isinstance(w_value, W_BaseException):
+                w_tb = w_value.w_traceback
+                if w_tb is None or space.is_w(w_tb, space.w_None):
```
cfbolz: this can be written as `space.is_none(w_tb)`, which checks for both None and w_None

cfbolz:
apart from my comment this looks good!

pypy/interpreter/pyopcode.py
StanFromIreland:
```suggestion
                if space.is_none(w_tb):
```
Thanks, I hadn't noticed it!

cfbolz:
thanks!

PR #5427, opened 2026-04-11
vstring: fix handle_str_equal_level2 not to getstrlen again

youknowone:
I found this while reading the code. Not sure it actually trigger bug or not.
Please feel free to close it if this is inappropriated. (but adding a comment why arg1 is justfied in this place will be great)

### DIFF
--- a/rpython/jit/metainterp/optimizeopt/vstring.py
+++ b/rpython/jit/metainterp/optimizeopt/vstring.py
@@ -717,10 +717,10 @@ class OptString(Optimization):
         handled, result = self.handle_str_equal_level1(arg2, arg1, op, mode)
         if handled:
             return True, result
-        handled, result = self.handle_str_equal_level2(arg1, arg2, op, mode)
+        handled, result = self.handle_str_equal_level2(arg1, arg2, l2box, op, mode)
         if handled:
             return True, result
-        handled, result = self.handle_str_equal_level2(arg2, arg1, op, mode)
+        handled, result = self.handle_str_equal_level2(arg2, arg1, l1box, op, mode)
         if handled:
             return True, result
         #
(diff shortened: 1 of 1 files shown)


mattip:
Would be nice to have a test that fails before, passes afterwards

youknowone:
While trying to create a test about it, I found it is a duplicated getstrlen call. The wrong argument was ignored because it was guaranteed to hit cache in that case. I couldn't create a test changed by the patch because it doesn't trigger bug. Rather than adding a test, I tried to remove the second cache. This is looking better if one more argument count is better than a getstrlen call.
Passing test shows getstrlen was not necessary there. If getstrlen is desired, probably putting neither arg1 nor arg2 will make sense not to confuse readers

youknowone:
@mattip Could you check it again?

cfbolz:
ok, I've read the code and understand it a bit more. your diff makes sense: don't call getstrlen a second time inside handle_str_equal_level2. so your change is an optimization, fine. I think the same change could be done to handle_str_equal_level1, right?

however, I don't understand why you think there is a bug here?

rpython/jit/metainterp/optimizeopt/vstring.py
```diff
@@ -786,12 +786,9 @@ def handle_str_equal_level1(self, arg1, arg2, resultop, mode):
         #
         return False, None
 
-    def handle_str_equal_level2(self, arg1, arg2, resultbox, mode):
+    def handle_str_equal_level2(self, arg1, arg2, l2box, resultbox, mode):
         i1 = getptrinfo(arg1)
         i2 = getptrinfo(arg2)
-        l2box = None
-        if i2:
-            l2box = i2.getstrlen(arg1, self, mode)
```
youknowone, line 794: @cfbolz because this line is referring arg1 for l2box. I expected arg2 to be aligned to other parts
Didnt know it is dead code atm

cfbolz: aaah, now I see it too :woman_facepalming: ok, makes perfect sense!

PR #5432, opened 2026-04-13
Lib python2

mattip:
Fix more lib-python failures.

Along the way we found a theoretical optimization in a `__call__` path, using a new `method_descriptor` flag on types. The flag is needed to properly set `__flags__` on types, which `lib-python.3.test.test_call` checks.


pypy/module/cpyext/object.py
```diff
@@ -510,6 +510,15 @@ def Py_Is(space, w_obj1, w_obj2):
...
 @cpython_api([], rffi.INT_real, error = -1)
 def PyGC_Collect(space):
+    from rpython.rlib.objectmodel import we_are_translated
+    from rpython.rlib import rawrefcount
+    from pypy.module.cpyext.state import _rawrefcount_perform
     import gc
-    gc.collect()
+    if not we_are_translated():
+        rawrefcount._collect()
+        space.user_del_action._run_finalizers()
```
cfbolz: it might make sense to run `_run_finalizers` from `interp_gc.py` here

pypy/module/cpyext/state.py
```diff
@@ -75,6 +75,10 @@ def check_and_raise_exception(self, always=False):
         operror = ec.cpyext_operror
         if operror:
             self.clear_exception()
+            try:
+                operror.normalize_exception(self.space)
+            except OperationError as e:
+                raise e
```
cfbolz: it looks like the try... except isn't needed here?

lib_pypy/_pypy_remote_debug.py
```diff
@@ -292,6 +292,8 @@ def write_memory(pid, address, content):
 
 
 def _read_and_parse_maps(pid='self', filter=None):
+    if filter is not None and os.path.isabs(filter):
+        filter = os.path.realpath(filter)
```
cfbolz, line 297: this condition sounds wrong to me? if it is an absolute path, realpath does nothing? is there a not missing?

mattip: The difference is to resolve symlinks. I will add a comment. This is failing on the buildbot since it is run in a venv.

mattip:
Comparing this branch to py3.11 with speed.pypy.org, it seems there is no real impact to the change in typeobject. Some benchmarks are ~10% faster, some ~5 slower, both with and without the JIT.

Issue #5048, opened 2024-09-24
refactor OperationError to just store a `w_value`


cfbolz:
- [x] "When an active exception is re-raised by a [raise](https://docs.python.org/3/reference/simple_stmts.html#raise) statement with no parameters, the traceback attached to this exception is now always sys.exc_info()[1].__traceback__. This means that changes made to the traceback in the current [except](https://docs.python.org/3/reference/compound_stmts.html#except) clause are reflected in the re-raised exception. (Contributed by Irit Katriel in [bpo-45711](https://bugs.python.org/issue?@action=redirect&bpo=45711).)"
- [x] "The interpreter state’s representation of handled exceptions (aka exc_info or _PyErr_StackItem) now only has the exc_value field; exc_type and exc_traceback have been removed, as they can be derived from exc_value. (Contributed by Irit Katriel in [bpo-45711](https://bugs.python.org/issue?@action=redirect&bpo=45711).)"

cfbolz:
I did a few of the changes related to this issue, such as making `PyErr_SetExcInfo` ignore the `py_type` and `py_traceback` arguments and take the information from the `py_value` argument.

The core change we didn't really have to do: pypy has always stored just a single `OperationError` to store the current exception. That could be changed some time to also not contain a `w_type`, but it's a refactoring for later. I'll leave the issue open for that.

mattip:
I think f2c8d4e closes this issue. I couldn't find anywhere else we relate to triples and exceptions.

mattip:
Closing as complete. Please reopen if I am mistaken.

PR #5433, opened 2026-04-13
Exceptiontable

### LINKED ISSUE #5041: 3.11: zero-cost exceptions
it's not clear that they are an optimization for pypy, but we should try it out. the behavior shouldn't change.

mattip:
Based on [a plan](https://gist.github.com/mattip/429841329e0eefdb1fb926ebb83c65f2) to port the CPython co_exceptiontable. Fixes #5041, related to #5048.


mattip:
No, this was the wrong set of commits.

PR #5434, opened 2026-04-13
Exceptiontable

### LINKED ISSUE #5041: 3.11: zero-cost exceptions

mattip:
Based on [a plan](https://gist.github.com/mattip/429841329e0eefdb1fb926ebb83c65f2) to port the CPython co_exceptiontable. Fixes #5041, related to #5048.


mattip:
These commits are for stages 0,1,2,3 of the plan

pypy/interpreter/pycode.py
```diff
@@ -223,6 +226,56 @@ def _cleanup_(self):
...
+            shift = 6
+            while b & 64:
+                b = ord(table[i]); i += 1
+                start |= (b & 63) << shift
+                shift += 6
+            start *= 2
+            b = ord(table[i]); i += 1
+            length = b & 63
+            shift = 6
+            while b & 64:
```
cfbolz: the decoding of the varsized ints in this function needs to be in a helper function, not duplicated several times. the encoding in the parser/decoding here needs a hypothesis round-tripping test

pypy/interpreter/pyopcode.py
```diff
@@ -118,6 +118,7 @@ def handle_generator_error(self, operr):
         ec = self.space.getexecutioncontext()
         return self.handle_operation_error(ec, operr)
 
+    @jit.dont_look_inside
```
cfbolz: this dont_look_inside is wrong. it's possible that we need it for now, but we need to add a TODO comment that we need to remove it in the end. exception paths are rare, but we do jit them regularly, and the generated code needs to be reasonable when traced.

pypy/interpreter/pycode.py
```diff
@@ -223,6 +226,56 @@ def _cleanup_(self):
     def signature(self):
         return self._signature
 
+    def lookup_exceptiontable(self, instr_offset):
```
cfbolz, line 230: this function should get a `@jit.elidable` decorator

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -665,11 +717,35 @@ def propagate_positions(self, blocks):
...
                 prev_position = instr.update_position_if_not_set(prev_position)
                 if instr.jump is not None and instr.jump.marked == 1 and instr.jump.instructions:
-                    if instr.opcode in (ops.SETUP_ASYNC_WITH, ops.SETUP_WITH, ops.SETUP_EXCEPT, ops.SETUP_FINALLY):
-                        continue # don't do this for exception handlers
+                    if instr.opcode in (_SETUP_WITH, _SETUP_FINALLY, _SETUP_CLEANUP):
+                        continue # don't propagate line info into exception handler blocks
                     instr.jump.instructions[0].update_position_if_not_set(prev_position)
-            if block.next_block and block.next_block.marked == 1 and block.instructions:
-                block.instructions[0].update_position_if_not_set(prev_position)
+            if (block.next_block and not block.cant_add_instructions and
```
cfbolz: nitpick, I would pull define a local variable `next_block = block.next_block` here

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -665,11 +717,35 @@ def propagate_positions(self, blocks):
...
+                    block.next_block.marked == 1 and block.next_block.instructions):
+                next_first_op = block.next_block.instructions[0].opcode
+                if next_first_op not in (_SETUP_CLEANUP, _SETUP_WITH, _SETUP_FINALLY):
+                    block.next_block.instructions[0].update_position_if_not_set(prev_position)
+
+    def guarantee_lineno_for_exits(self, blocks):
+        # CPython equivalent: guarantee_lineno_for_exits (compile.c 8278-8299).
```
cfbolz: let's remove CPython line numbers, those are bound to change over time.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -721,11 +798,15 @@ def _finalize_blocks(self):
             self.emit_op(ops.RETURN_VALUE)
             self.current_block.auto_inserted_return = True
         blocks = self.first_block.post_order()
+
         remove_redundant_nops(blocks)
         self.jump_thread(blocks)
+        self.extend_blocks(blocks)
+        apply_static_swaps(blocks)
```
cfbolz: apply_static_swaps should be a method on PythonCodeMaker, to be in the same style as all the other methods called from _finalize_blocks

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+          depth_adjust = _opcode_stack_effect_jump(op) for the opening SETUP_*.
+          depth_sub = 1 for _SETUP_WITH (__enter__ result consumed before body), 0 otherwise.
+        """
+        # --- Phase 1: graph traversal, compute per-block entry handler stacks ---
+        # Like CPython's label_exception_targets: DFS over the CFG, tracking a
+        # stack of active handler blocks.  stk is list of Block (not tuples);
+        # per-handler metadata (lasti, depth_adj, depth_sub) is stored as
+        # attributes on the handler Block itself (CPython: b_lasti, b_exceptdepth).
+        visited = {}           # Block -> True
+        block_entry_stack = {} # Block -> list of Block
```
cfbolz, line 865: do we really need both of these dicts? isn't `block_entry_stack` only filled if the block is in visited? if that is true, let's remove `visited`.

If we absolutely need both, I think the use of a hash-based set for visited is inefficient here. can we instead use the general purpose `.marked` attribute of `Block` for visited information?

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+                        visited[target] = True
+                        block_entry_stack[target] = stk[:]
+                        worklist.append(target)
+
+            # Propagate to fallthrough successor.
+            # CPython's label_exception_targets always follows b_next regardless
+            # of whether the block ends with an unconditional jump.  We must do
+            # the same: after extend_blocks/jump_thread removes a JUMP_FORWARD,
+            # cant_add_instructions is still True but next_block still needs
+            # handler-stack propagation.
```
cfbolz: should we set `cant_add_instructions` to `False` if we extend blocks and remove the `JUMP_FORWARD`? it feels weird to have this outdated value in there.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+        block_entry_stack[first_block] = []
+        worklist = [first_block]
+
+        while worklist:
+            b = worklist.pop()
+            stk = block_entry_stack[b][:]  # mutable working copy (list of Block)
+
+            for instr in b.instructions:
+                op = instr.opcode
+                if op in (_SETUP_FINALLY, _SETUP_CLEANUP, _SETUP_WITH):
```
cfbolz: there are lots of different ways that this condition is spelled out in this file. let's define a method `Instruction.is_setup_pseudo_opcode` to do it and call that everywhere

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+                        visited[handler_target] = True
+                        block_entry_stack[handler_target] = stk[:]
+                        worklist.append(handler_target)
+                    # Record handler metadata on the block (like CPython b_lasti).
+                    handler_target.exc_lasti = (op != _SETUP_FINALLY)
+                    handler_target.exc_depth_adj = _opcode_stack_effect_jump(op)
+                    handler_target.exc_depth_sub = 1 if op == _SETUP_WITH else 0
+                    stk.append(handler_target)
+                elif op == _POP_BLOCK:
+                    if stk:
```
cfbolz, line 889: when can `stk` be empty here? shouldn't that be an `assert stk` instead? if it can be empty, please add a comment when that happens.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+                        worklist.append(target)
+
+            # Propagate to fallthrough successor.
+            # CPython's label_exception_targets always follows b_next regardless
+            # of whether the block ends with an unconditional jump.  We must do
+            # the same: after extend_blocks/jump_thread removes a JUMP_FORWARD,
+            # cant_add_instructions is still True but next_block still needs
+            # handler-stack propagation.
+            if b.next_block is not None:
+                if b.next_block not in visited:
```
cfbolz: we can express this as one if and remove a layer of nesting: `if b.next_block is not None and b.next_block not in visited`

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -734,6 +815,156 @@ def _finalize_blocks(self):
...
+        all_entries = []
+        cur_handler = None   # type: Block or None
+        cur_start = -1
+        instr_offset = 0
+
+        for block in blocks:
+            entry = block_entry_stack.get(block)
+            if entry is None:
+                # Dead block: close any open range.
+                if cur_handler is not None and cur_start < block.offset:
```
cfbolz: this pattern `if curr_handler ...: all_entries.append((...))` appears three times in this function. let's factor it out into a small helper method (or a local function, but careful, rpython doesn't support closures capturing local variables. you can still define one if you pass in as arguments everything you need).

pypy/interpreter/astcompiler/test/test_assemble.py
```diff
@@ -69,6 +74,31 @@ def test_remove_redundant_nops():
...
+    result = []
+    for v in values:
+        _encode_varint(result, v)
+    table = ''.join(result)
+    i = 0
+    for v in values:
+        decoded, i = _decode_varint(table, i)
+        assert decoded == v
+    assert i == len(table)
+
```
cfbolz, line 313: there needs to be an extra hypothesis test that checks the properties of passing the msb argument to _encode_varint.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -768,16 +1000,63 @@ def assemble(self):
...
+            if target is None or not target.instructions:
+                continue
+            if len(target.instructions) > MAX_COPY_SIZE:
+                continue
+            has_return = False
+            for instr in target.instructions:
+                if instr.opcode == ops.RETURN_VALUE:
+                    has_return = True
+                    break
+            if not has_return:
```
cfbolz: can't we use an else block in the for loop here, instead of the `has_return` bool?

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -768,16 +1000,63 @@ def assemble(self):
...
+                continue
+            has_lineno = False
+            for instr in target.instructions:
+                if instr.position_info[0] != -1:
+                    has_lineno = True
+                    break
+            if has_lineno:
+                continue
```
cfbolz, line 1049: use an else block after the for loop here as well

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -825,8 +1104,25 @@ def should_mark_block(block):
...
-                # if it's an unconditional jump, duplicate it
-                if op.opcode in (ops.JUMP_FORWARD, ops.JUMP_ABSOLUTE):
+                # if it's an unconditional jump, inline it (single predecessor)
+                # or copy it (multiple predecessors, matching CPython behaviour).
+                if op.opcode in (ops.JUMP_FORWARD, ops.JUMP_ABSOLUTE) and (target.marked >> 1) <= 1:
+                    # Skip if target is a RERAISE-only block: duplicating it
+                    # INTO this block would place the RERAISE inside the byte
+                    # range of an enclosing try-body's exception-table entry
+                    # (body blocks emitted by _visit_try_except/_visit_try_finally
+                    # have a table entry body.offset..exc.offset, and the
```
cfbolz: double `..` typo?

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -825,8 +1104,25 @@ def should_mark_block(block):
...
+                    # range, dispatching to the inner handler instead of the
+                    # outer reraise cleanup).  Mirrors the RERAISE-only skip
+                    # in the other branch below.
+                    reraise_only = bool(target.instructions)
+                    for _instr in target.instructions:
+                        if _instr.opcode != ops.RERAISE:
+                            reraise_only = False
+                            break
+                    if reraise_only:
+                        continue
```
cfbolz, line 1144: I find the way to check whether the target is only a reraise very confusing. why is there a loop? shouldn't this simply be checking that there is exactly one instruction, and that is a RERAISE?

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -847,6 +1143,23 @@ def should_mark_block(block):
...
+                    # Leaving the original block in place keeps it within range.
+                    # Position accuracy is unaffected: RERAISE uses attach_tb=False
+                    # so no traceback entry is ever generated at the RERAISE site.
+                    reraise_only = True
+                    for _instr in target.instructions:
+                        if _instr.opcode != ops.RERAISE:
+                            reraise_only = False
+                            break
+                    if reraise_only:
+                        continue
```
cfbolz, line 1177: logic duplication with the reraise_only above. factor it out into a helper function.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -863,7 +1176,13 @@ def should_mark_block(block):
...
                         newtarget.emit_instr(instr)
                     op.jump = newtarget
-                    blocks.append(newtarget)
+                    # Match CPython: new_target->b_next = target->b_next; target->b_next = new_target
+                    for k in range(len(blocks)):
+                        if blocks[k] is target:
+                            blocks.insert(k + 1, newtarget)
+                            break
+                    else:
+                        blocks.append(newtarget)
```
cfbolz: ouch, this insert makes `duplicate_exits_without_lineno` potentially quadratic in the number of block, doesn't it?

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -1002,6 +1345,110 @@ def remove_redundant_nops(blocks):
...
+                store_idx = _stores_to(instructions[idx])
+                if store_idx >= 0 and (store_idx == store_j or store_idx == store_k):
+                    return
+        # success: replace SWAP with NOP and swap the two instructions
+        instr.opcode = ops.NOP
+        instructions[j], instructions[k] = instructions[k], instructions[j]
+        i -= 1
+
+
+def apply_static_swaps(blocks):
```
cfbolz: I would like explicit unit tests for apply_static_swaps. create some blocks with manually created instructions (ie not using the rest of the ast compiler at all). call apply_static_swaps. check the result. the tests should cover all the cases.

pypy/interpreter/astcompiler/assemble.py
```diff
@@ -1231,6 +1683,8 @@ def func(arg, _value=value):
 
 def _opcode_stack_effect(op, arg):
     """Return the stack effect of a opcode an its argument."""
+    if op == _SETUP_FINALLY or op == _SETUP_CLEANUP or op == _SETUP_WITH or op == _POP_BLOCK:
```
cfbolz: there is `is_pseudo_opcode` for this condition.

cfbolz (commented):
I've reviewed assemble.py only so far, will continue on the rest later.

pypy/interpreter/pyframe.py
```diff
@@ -75,6 +76,14 @@ class PyFrame(W_Root):
...
     w_yielding_from = None
     last_instr               = -1
+    # Transient field: set by RERAISE N (N > 0) to the lasti value it read
+    # from the stack, so that handle_operation_error can push that as the
+    # new handler's lasti (preserving the original raise site across
+    # re-raises).  Cleared by handle_operation_error after use.  -1 means
+    # "not set" -- use intmask(self.last_instr) instead.  Mirrors CPython's
+    # trick of setting frame->prev_instr = first_instr + lasti in RERAISE
+    # and then having exception_unwind read _PyInterpreterFrame_LASTI.
+    _reraise_lasti           = -1
```
cfbolz: this field is expensive. it makes Frame objects one word larger, and allocating frames is a pretty significant part of the memory allocations that the interpreter does. can we discuss strategies for not needing that field?

pypy/interpreter/pyframe.py
```diff
@@ -878,126 +903,219 @@ def first_line_not_before(lines, line):
...
-    elif kind == JUMP_BLOCKSTACK_WITH:
+    elif kind == _MS_WITH:
         return "can't jump into the body of a with statement"
     else:
-        assert kind == JUMP_BLOCKSTACK_EXCEPT
-        return "can't jump into an 'except' block as there's no exception"
+        return "can't jump to target"
 # ____________________________________________________________
 
 def get_block_class(opname):
```
cfbolz: I think get_block_class and unpickle_block are now unused

pypy/interpreter/pyframe.py
```diff
@@ -863,11 +862,37 @@ def descr_repr(self, space):
         return self.getrepr(space, "frame", moreinfo)
 
 # ____________________________________________________________
-
-JUMP_BLOCKSTACK_WITH = 'w'
-JUMP_BLOCKSTACK_LOOP = 'l'
-JUMP_BLOCKSTACK_TRY = 't'
-JUMP_BLOCKSTACK_EXCEPT = 'e'
+# Abstract value-stack kind constants for mark_stacks / compatible_stack.
+# Stack is packed into a Python int: 3 bits per slot, TOS in lowest bits.
```
cfbolz: the comment should at least mention that this is for fset_f_lineno.

pypy/interpreter/pyframe.py
```diff
@@ -878,126 +903,219 @@ def first_line_not_before(lines, line):
...
+        return True
+    return False
+
+def mark_stacks(code):
+    """Compute abstract value-stack state at each instruction index.
+
+    Returns a list of length (len(co_code)//2 + 1).  Each entry is either
+    _MS_UNINITIALIZED, _MS_OVERFLOWED, or a non-negative int whose 3-bit
+    slots encode the kind of each value-stack slot (TOS in lowest bits).
+    """
```
cfbolz, line 976: right now, we have good integration tests for mark_stacks in the form of actually setting the lineno in a frame from a trace function. however, the logic is really complicated, and I would like to have some unit tests that manually constructs at least a few bytecode objects, calls mark_stacks and asserts that the result is as expected.

pypy/interpreter/pyframe.py
```diff
@@ -878,126 +903,219 @@ def first_line_not_before(lines, line):
...
+            lasti = dl & 1
+            start_stack = stacks[start_raw]
+            if start_stack != _MS_UNINITIALIZED:
+                handler_stack = _ms_pop_to_level(start_stack, depth)
+                # CPython-parity dispatch: if lasti, push the raise-site
+                # offset int before pushing the exception.
+                if lasti:
+                    handler_stack = _ms_push(handler_stack, _MS_LASTI)
+                handler_stack = _ms_push(handler_stack, _MS_EXCEPT)
+                if stacks[target_raw] == _MS_UNINITIALIZED:
```
cfbolz: why can't _ms_set be used here?

pypy/interpreter/pyframe.py
```diff
@@ -863,11 +862,37 @@ def descr_repr(self, space):
...
+_MS_EXCEPT        = 2   # prev_exc pushed by PUSH_EXC_INFO
+_MS_OBJECT        = 3   # ordinary value
+_MS_WITH          = 4   # __exit__ callable pushed by BEFORE_WITH (PyPy-specific)
+_MS_LASTI         = 5   # raise-site offset pushed on lasti=True handler dispatch
+_MS_WILL_OVERFLOW = 1 << (20 * _MS_BITS)
+
+def _ms_push(stack, kind):
+    if stack < 0 or stack >= _MS_WILL_OVERFLOW:
+        return _MS_OVERFLOWED
+    return (stack << _MS_BITS) | kind
```
cfbolz: I think we should use the fact that rpython has classes here. please abstract the mark stack logic into a new class `MarkStacks`, which is also what `mark_stacks` should return. make the `_ms_*` helpers methods on that class. write some unit tests for the class too.

pypy/interpreter/pyframe.py
```diff
@@ -878,126 +903,219 @@ def first_line_not_before(lines, line):
...
-            ):
-                if i + 2 < len(code.co_code):
-                    next_opcode = ord(code.co_code[i + 2])
-                else:
-                    next_opcode = NOP
-                if next_opcode != CALL_FUNCTION:
-                    block_stack = block_stack + JUMP_BLOCKSTACK_LOOP
-                blocks[i // 2 + 1] = block_stack
+                j = arg + i + 1   # instruction index
+                if _ms_set(stacks, j, stack):
```
cfbolz: I think the pattern `if _ms_set(...): changed = True` is unnecessarily verbose. let's write this as `changed = _ms_set(stacks, j, stack, changed)` (change `_ms_set` accordingly).

pypy/interpreter/pyopcode.py
```diff
@@ -147,20 +147,49 @@ def handle_operation_error(self, ec, operr, attach_tb=True):
...
-            unroller = SApplicationException(operr)
-            next_instr = block.handle(self, unroller)
-            return next_instr
+        entry = self.getcode().lookup_exceptiontable(self.last_instr)
+        target, depth, lasti = entry
+        if depth >= 0:
+            # depth is relative (0 = empty value stack); convert to absolute.
+            code = self.getcode()
+            stackstart = (code.co_nlocals + len(code.co_cellvars) +
+                          len(code.co_freevars))
```
cfbolz: introduce a helper method `_stack_start()` on PyFrame and use it. also use it in `_check_stack_index` and `PyFrame.__init__`

pypy/interpreter/pyopcode.py
```diff
@@ -1298,47 +1330,70 @@ def SETUP_WITH(self, offsettoend, next_instr):
...
+        # avoids a spurious traceback entry and exception_trace call that
+        # would occur if the OperationError bubbled up to the plain
+        # 'except OperationError' branch in handle_bytecode.
+        raise RaiseWithExplicitTraceback(operr)
 
     def CALL_FUNCTION(self, oparg, next_instr):
         # Only positional arguments
         nargs = oparg & 0xff
         w_function = self.peekvalue(nargs)
+        if isinstance(w_function, function.Function):
```
cfbolz: the changes to CALL_FUNCTION look incorrect. why are they there? `space.call_valuestack` contains more logic and has the same fast path. if this is needed for correctness reasons, please add a comment why. otherwise, remove?

pypy/module/_rawffi/test/test__rawffi.py
```diff
@@ -826,6 +826,19 @@ def callback():
             assert res[0] == 0
         finally:
             sys.stderr = orig
+        # DEBUG: dump exception tables of the test fn and the inner callback
```
cfbolz: is the DEBUG logic still needed?

cfbolz (commented):
done reviewing

pypy/interpreter/astcompiler/assemble.py
mattip: good catch, we can get rid of visited

pypy/interpreter/astcompiler/assemble.py
mattip: added missing update of cant_add_instructions in optimize_unreachable_code, changed the comment to only reference the CPython behaviour.

pypy/interpreter/astcompiler/assemble.py
mattip: cannot be empty

pypy/interpreter/astcompiler/assemble.py
mattip: this one is backwards: if we __did__ find a lineno (i.e. early break from the loop) we continue.

pypy/interpreter/astcompiler/assemble.py
mattip: Refactored to mimic a double-linked list like CPython. Might need another cycle of review.

pypy/interpreter/pyopcode.py
mattip: Commit a21646207b9 is not part of the exceptiontable branch, it is part of some other work. I will revert it here.

mattip:
I related to most of the review comments. The `pyopcode.py` comments still need to be related to. I have an idea how to avoid adding the attribute to Frame, and will add more tests.

pypy/interpreter/pyframe.py
mattip: Worked around it by adding the field to `RaiseWithExplicitTraceback` instead

mattip:
[Speed comparison](https://speed.pypy.org/comparison/?exe=22%2BL%2Bpy3.11%2C22%2BL%2Bexceptiontable&ben=1%2C34%2C58%2C63%2C60%2C27%2C61%2C2%2C25%2C57%2C3%2C46%2C4%2C5%2C41%2C42%2C22%2C44%2C6%2C59%2C39%2C7%2C8%2C65%2C45%2C23%2C62%2C66%2C24%2C9%2C10%2C47%2C48%2C49%2C50%2C51%2C11%2C12%2C13%2C40%2C14%2C69%2C15%2C70%2C67%2C68%2C64%2C35%2C36%2C37%2C38%2C16%2C52%2C54%2C55%2C53%2C56%2C28%2C30%2C32%2C29%2C33%2C17%2C18%2C19%2C20%2C43&env=3&hor=true&bas=22%2BL%2Bpy3.11&chart=normal+bars) looks ok. [buildbot](https://buildbot.pypy.org/summary?branch=exceptiontable) too.

pypy/interpreter/astcompiler/assemble.py
cfbolz: ah, it was supposed to be a range, I'll add it back

pypy/interpreter/pyframe.py
cfbolz: this was a really excellent idea, looks great!

cfbolz:
I fixed a few minor nits. I am happy now, I think. well done @mattip! feel free to merge

PR #5442, opened 2026-04-16
redo release_buffer branch

mattip:
Redo release_buffer from PR #5407. Now that py3.11 is almost clear of failures it will be easier to see if this is correct.


mattip:
Closing in favor of #5444

PR #5444, opened 2026-04-18
Use a rpython Buffer like CPython's Py_buffer

mattip:
Make PyPy's buffer handling 1:1 like CPython's buffer. A rework of #5442


mattip:
There is a failing lib-python/3/test/test_re test (unmodified) on pypy3.11-v7.3.21. Fixing it required adding a way to manage buffer lifetime. This led me to look at our `BufferView` hierarchy versus CPython's `Py_buffer`. It turns out we could do better — and fix a whole class of buffer lifetime problems in cpyext and hpy — by adopting the CPython model. A `Py_buffer` is now a lightweight passive struct that lives alongside the existing `BufferView` hierarchy. Each view is an independent description of the same underlying data, but all views share the same exporter. Since `W_BytearrayObject` is mutable, it carries an `_exports` counter (introduced in commit 938bc20): while it is positive the object cannot be mutated or resized. `W_MemoryView` and `W_PickleBuffer` register a finalizer to decrement the counter if `release()` is never called explicitly. New `needs_release()` and `owns_export` flags let `newmemoryview()` skip the finalizer for immutable exporters like bytes, keeping those views JIT-virtualizable.

I ran speed.pypy.org on an earlier version of these commits and saw no regressions. I will run benchmarks and the buildbot again on this set of reworked commits.

mattip:
The buildbot shows this branch fixes the lib-python test_re failure. It would be difficult to fix that without this approach. The speed.pypy.org comparison shows it does not significantly change performance for our benchmarks.

mattip:
@cfbolz any thoughts?

pypy/interpreter/pyopcode.py
```diff
@@ -2011,6 +2011,7 @@ def source_as_str(space, w_source, funcname, what, flags):
             raise oefmt(space.w_TypeError,
                         "%s() arg 1 must be a %s object", funcname, what)
         source = buf.as_str()
+        buf.releasebuffer()
```
cfbolz: shouldn't this be in a try finally?

pypy/module/__pypy__/interp_buffer.py
```diff
@@ -20,7 +21,10 @@ def descr_buffer(self, space, w_flags):
 
     def readbuf_w(self, space):
         mv = space.call_method(self, '__buffer__', space.newint(0))
-        return mv.buffer_w(space, 0).as_readbuf()
+        buf = mv.buffer_w(space, 0)
+        result = buf.as_readbuf()
+        buf.releasebuffer()
```
cfbolz: try finally?

pypy/module/_io/interp_bufferedio.py
```diff
@@ -118,22 +118,25 @@ def readinto1_w(self, space, w_buffer):
...
+            data = space.bytes_w(w_data)
+            if len(data) > length:
+                raise oefmt(space.w_ValueError,
+                            "%s() returned too much data: "
+                            "%d bytes requested, %d returned",
+                            methodname, length, len(data))
+            self.output_slice(space, rwbuffer, 0, data)
+            return space.newint(len(data))
+        finally:
+            view.releasebuffer()
```
cfbolz: looking at this, I am wondering whether we could have a context manager API that's responsible for releasing the views? that might make it harder to get the use of buffers wrong.

mattip: created a context manager and applied it anywhere I could

mattip:
I ran the buildbot on ee95f72 and fixed the failure from a bad merge in HEAD. I am 99% sure the buildbot failures are all ones that are not related to this branch. The speed.pypy.org comparison to py3.11 is noisy, but I don't think this causes a slowdown.

pypy/module/cpyext/unicodeobject.py
```diff
@@ -866,7 +866,8 @@ def PyUnicode_FSDecoder(space, w_obj, result):
                 "path should be %s, not %s" % (allowed_types, tp,)),
                 space.w_DeprecationWarning)
             buffer = space.buffer_w(w_obj, space.BUF_FULL_RO)
-            w_output =  space.fsdecode(space.newbytes(buffer.as_str()))
+            with buffer:
```
cfbolz, line 869: I don't want to be too annoying, but wouldn't this kind of code be more idiomatic?:

```python
with space.buffer_w(w_obj, space.BUF_FULL_RO) as buffer:
    w_output = space.fsdecode(space.newbytes(buffer.as_str()))
```

feel free to ignore this comment, it would require yet another pass over this PR :-(

cfbolz:
apart from my last comment I think we can merge

PR #5447, opened 2026-04-19
Refactor w type use in OperationError

mattip:
Related to #5048. It converts `w_type` to `_w_type`, which is non-`None` at OperationError setup and `None` after error normalization. To truly eliminate the field, we would need to eagerly normalize exceptions which means adding a `space` argument to the oefmt/OperationError calls, which is a bigger change.


PLAN_remove_w_type.md
StanFromIreland, line 1: Um, I don't think we want to add this to the repo? I suggest you add `.claude/` to the .gitignore (we did this in CPython), and have it create such things in there.

mattip: Done, I meant to remove the plan once the branch is reviewed but it provides a bit of context for the changes.

PR #5451, opened 2026-04-26
Update all docs links after `/en/latest/` prefix was removed

StanFromIreland:
I also slightly rewrote `pypy/module/README.txt` a long the way since it was a tad messy.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,7 +36,7 @@ Building PyPy is not the recommended way to obtain the PyPy alternative python
 interpreter. It is time-consuming and requires significant computing resources.
 More information can be found here:
 
-    https://doc.pypy.org/en/latest/build.html
+    https://doc.pypy.org/build.html
 
 Enjoy and send us feedback!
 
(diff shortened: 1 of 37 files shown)


Carreau:
moving and removing /en/ also broke all intersphinx, and all project having inter sphinx with pypy in their conf.py

StanFromIreland:
Indeed, there should be a redirect, but I don’t know if it was set up.

Carreau:
I also opened an issue on pypy.org, https://github.com/pypy/pypy.org/issues/178 I'll update intersphinx registry to point to the new url.

mattip:
> Indeed, there should be a redirect, but I don’t know if it was set up.

No, I did not do a redirect. I need to do it at the DNS level or can I do it on readthedocs?

StanFromIreland:
You can do it in Readthedocs. I sent an email with a link to the docs for it.

mattip:
Ahh, found it. I added the top-level redirect so the old links should still work

PR #5453, opened 2026-04-29
Fix some typos spotted by lintian

stefanor:
These ones are specific to the py3.11 branch, and build on top of #5452.


mattip:
Thanks for both the PRs.

PR #5457, opened 2026-04-29
small optimizations

mattip:
I was using perf and callgrind to examine a toy benchmark around docutils and polymorphic classes in the bare interpreter without the JIT. A few things jumped out:
- ~every frame checks stack overflow. The call is not inlined. I refactored the way the function is wrapped so it could be inlined.~
- calls to `func.__name__`, `func.__qual_name__` and `type.__name__` regenerate a `W_Unicode` object which should be inexpensive but instead calls `codepoints_in_utf8` to figure out the length each time. Caching these avoids the calls. This is heavily used in docutils. It is a bit of a micro-optimization to make the sphinx benchmark faster, but I think those attributes are frequently used elsewhere too.
- a simple refactor to avoid unneeded length calculation in a fast path

### DIFF
--- a/pypy/interpreter/function.py
+++ b/pypy/interpreter/function.py
@@ -45,6 +45,8 @@ class Function(W_Root):
     w_objclass = None
     w_text_signature = None
     _empty_defs = []
+    _w_name = None
+    _w_qualname = None
 
     def __init__(self, space, code, w_globals=None, defs_w=[], kw_defs_w=None,
                  closure=None, w_ann=None, forcename=None, qualname=None):
@@ -362,7 +364,9 @@ class Function(W_Root):
 
         self.space = space
         self.name = space.text_w(w_name)
+        self._w_name = w_name
         self.qualname = space.utf8_w(w_qualname)
+        self._w_qualname = w_qualname
         self.code = space.interp_w(Code, w_code)
         if not space.is_w(w_closure, space.w_None):
             from pypy.interpreter.nestedscope import Cell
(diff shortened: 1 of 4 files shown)


cfbolz:
The non stack check look good.

The motivation for the stack check changes is very sensible too, but I'm worried about some of the details. Could you make a separate PR for them (and also target main)?

mattip:
While the changes look like they should speed things up, the [comparison to latest HEAD without the JIT](https://speed.pypy.org/comparison/?exe=21%2BL%2Bpy3.11%2C21%2BL%2Bstack-overflow&ben=1%2C34%2C58%2C63%2C60%2C27%2C61%2C2%2C25%2C57%2C3%2C46%2C4%2C5%2C41%2C42%2C22%2C44%2C6%2C59%2C39%2C7%2C8%2C65%2C45%2C23%2C62%2C66%2C24%2C9%2C10%2C47%2C48%2C49%2C50%2C51%2C11%2C12%2C13%2C40%2C14%2C69%2C15%2C70%2C67%2C68%2C64%2C35%2C36%2C37%2C38%2C16%2C52%2C54%2C55%2C53%2C56%2C28%2C30%2C32%2C29%2C33%2C17%2C18%2C19%2C20%2C43&env=3&hor=true&bas=21%2BL%2Bpy3.11&chart=normal+bars) does not really show much improvement. Making this draft until I can get some better speedups.

mattip:
I split out the stack-check changes

PR #5460, opened 2026-04-30
inline fastpath for stackoverflow check

mattip:
Broken out of #5457.

I was using perf and callgrind to examine a toy benchmark around docutils and polymorphic classes in the bare interpreter without the JIT. I noticed that the call to check stack size is not inlined. I refactored the way the function is wrapped so it could be inlined.

### DIFF
--- a/rpython/translator/c/src/stack.h
+++ b/rpython/translator/c/src/stack.h
@@ -36,6 +36,19 @@ void LL_stack_set_length_fraction(double);
 /* some macros referenced from rpython.rlib.rstack */
 #define LL_stack_get_end() ((Signed)rpy_stacktoobig.stack_end)
 #define LL_stack_get_length() rpy_stacktoobig.stack_length
+
+/* Inline fast-path for the stack overflow check.  The generated C calls
+   pypy_g_stack_check() which lives in a separate compilation unit, so GCC
+   cannot inline it without LTO.  By putting the fast path here as a static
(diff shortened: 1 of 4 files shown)


rpython/translator/c/src/stack.h
```diff
@@ -36,6 +36,19 @@ void LL_stack_set_length_fraction(double);
...
+   cannot inline it without LTO.  By putting the fast path here as a static
+   inline, every translation unit gets the check inlined directly, avoiding
+   the function-call and SSP-canary overhead at each of the ~2600 call sites.
+   The slow path (LL_stack_too_big_slowpath) stays out-of-line. */
+static inline void RPY_STACK_CHECK(void) {
+    Signed current = (Signed)&current;
+    if ((size_t)(rpy_stacktoobig.stack_end - (char *)current) >
+            (size_t)rpy_stacktoobig.stack_length)
+        LL_stack_too_big_slowpath(current);
+}
```
mattip, line 51: This part is the important thing for this PR: to split out a fastpath that will hopefully be marked by the branch predictor and avoid the slow check entirely. Inlining is a precondition, the split is the important addition.

mattip:
~I think the [fixed branch](https://github.com/pypy/pypy/tree/stackcheck-explicit-inline) does not have the fastpath.~

mattip:
Closing in favor of #5471

PR #5465, opened 2026-05-05
Move implementation of `str.splitlines` to a helper for the JIT

StanFromIreland:
Carl's idea, it allows the JIT to unbox the wrapped string if it saw the allocation earlier in the trace. This is already done for `split(()`.


mattip:
Thanks @StanFromIreland

PR #5467, opened 2026-05-07
Py3.11 heapq tournament tree

### LINKED ISSUE #5466: heapq.merge slowdown from 3.10 to 3.11
Mostly wondering whether this is already known, and/or intentional.

I have a very merge-heavy program (generate all the divisors of the order of the "monster group" in increasing order - over 400 million(!) of 'em).  Running under Win10 64-bit:

pypy3.10-v7.3.12 145 seconds
pypy3.11-v7.3.21 394 seconds

The latter appears to have CPython's version of heapq.py, using heaps, while the former's heapq.py has Dennis Sweeney's algorithm using a tournament tree variant.  The former wins big on two counts:

- Fewer compares needed, in bad cases half as many.
- Elements are compared directly. The heap-based version needs to wrap elements in tuples, partly to inject a unique integer "tie breaker", to preserve stability (the tree version ensures that by construction). Slows comparisons and bloats the number of temp objects created.

I'm not clear on why Dennis's version was rejected for CPython - it works great for every use I've had. Or is there a flaw I never bumped into?

cfbolz:
fix #5466 by reintroducing the linked tournament tree-based heapq.merge. Thanks @sweeneyde for the original implementation and @tim-one for noticing that it had gotten lost!

### DIFF
--- a/extra_tests/test_heapq.py
+++ b/extra_tests/test_heapq.py
@@ -39,3 +39,7 @@ def test_merge_tree_sizes(random, n):
         for i in range(n)
     ]
     check_merge_case(lists)
+
+def test_check_linked_tournament_tree_in_place():
+    import heapq
+    assert hasattr(heapq, '_MergeNode')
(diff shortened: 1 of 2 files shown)


StanFromIreland (approved):
LGTM

PR #5468, opened 2026-05-08
Int ovf to long without fromint

cfbolz:
related to #5466: don't call `rbigint.fromint` for add, sub, lshift that overflow any more. instead, just compute the correct rbigint directly. we already use this approach for mul.


mattip:
I kicked of a [buildbot run](https://buildbot.pypy.org/builders/pypy-c-jit-linux-x86-64/builds/11749). Anything else we should check before this goes in?

PR #5469, opened 2026-05-10
turn 'app level lst = [None] * n' into a malloc + memset

mattip:
Looking at `PyFrame.__init__`, which is called many times. It has https://github.com/pypy/pypy/blob/220498e6f4c2a101923015df4903c22edca33b56/pypy/interpreter/pyframe.py#L109

which is not optimized to `raw_memclear` which becomes `OP_RAW_MEMCLEAR` in C which is `memset`. This PR optimizes it. The change is not really noticeable, but it is something.


cfbolz:
Maybe we can have a synthetic RPython program where we construct a really large array and try to see whether we can measure a difference there.

mattip:
Translating and testing a small rpython  program was a good idea. It turns out there are a few scenarios, and each interacts with the incminmark garbage collector differently. The pattern is that new gc arrays are allocated on the nursery and then the pointers zeroed out, so there are different modes for assignment of the form `lst = [thing] * size`:

n | Mode | Possible Optimization in this PR
--|--|--
0 | `[None]*n`  | Skip `ll_arrayclear` for GcPtr arrays — `zero_gc_pointers_inside` already handles it; the commit introduced a redundant `memset`                                                              
1 | `[0]*n` | None — `ll_arrayclear` is correct and necessary; GCC auto-vectorizes anyway                               
2 | `[obj]*n` | Replace `ll_fixed_setitem_fast` loop with `bare_setarrayitem` — eliminate unnecessary write barrier checks                                                                   
3 | loop `setitem`| None — write barriers required

So I added tests for all 4 scenarios and a benchmark, and fixed the code to make the tests pass. Translating the benchmark on the commit just before these changes and then on the changes, and running through callgrind to get total instruction count gave me this.

Target: `rpython/translator/goal/targetlistalloc.py`, 500 repeats over sizes [100, 1000, 10000].
Translation: <pypy2> rpython/bin/rpyhon -O2 --gc=incminimark <target>
Measurement: `valgrind --tool=callgrind <binary>`, then look for `Collected:` which is the total number of instructions

| Mode | Before `e070713120` | After `71f8941251` | Change |
|------|--------------------|--------------------|--------|
| 0 `[None]*n` — GcPtr, null item | 88,661,576 | 44,591,026 | **~2× faster** |
| 1 `[0]*n` — Signed, zero item | 44,530,337 | 44,530,266 | no change |
| 2 `[obj]*n` — GcPtr, non-null item | 64,103,072 | 64,095,063 | ~0 |
| 3 loop setitem — `lst[i]=obj` | 25,574,999 | 20,508,433 | ~20% |

The drop in mode 0 is the one I wanted, it is the pattern in pyframe which is a very hot code path.

mattip:
The last commit fixes tests but restores mode 3 to the pre-PR instruction count. The problem is the simulated GC in tests, it does not wrap bare_setitemarray. So dfcbbe7 prefers correct testability over a small optimization.

mattip:
FWIW, with the commit previous to dfcbbe7, many of the `test_transformed_gc.py` tests fail.

mattip:
Adding a single write barrier seems to have fixed the problem. I reran the benchmark, the result on 2b614c1 is the same as "After 71f8941251" above.

mattip:
There were failing tests in rpython/memory but 193491a fixed them.

rpython/rlib/rgc.py
```diff
@@ -508,26 +508,55 @@ def ll_shrink_array(p, smallerlength):
...
+def ll_arrayfill(p, item):
+    from rpython.rtyper.lltypesystem.lloperation import llop
+    from rpython.rlib.objectmodel import keepalive_until_here
+    i = 0
+    length = len(p)
+    if we_are_translated():
+        # One write_barrier before the loop covers all slots: if p is old,
+        # it gets added to old_objects_pointing_to_young so the minor GC will
+        # trace its nursery slots; if p is young, the barrier is a no-op.
+        llop.gc_writebarrier(lltype.Void, p)
```
cfbolz, line 549: so the reason I am wary about this PR is actually this line here. if the arrays are big, they use card marking. I don't understand things enough to be sure that this is correct in the presence of card marking arrays. I'll try to understand the details around card marking some more to decide whether this is fine or not. I'll also add arrayfill to our random gc hypothesis testing and push that.

cfbolz:
I looked through the code some more, I think it's fine. the test run came back clean too. I'm merging

PR #5474, opened 2026-05-16
Estimate GC nursery size via performance-core L2 cache (not efficiency) on Apple Silicon

Julian:
On Apple Silicon, `hw.l2cachesize` returns the *efficiency* cluster's L2 (e.g. 4 MB on a typical M-series), not the performance cluster's (16 MB and up). The result was that `get_L2cache_darwin()` returned 4 MB, which in `best_nursery_size_for_L2cache()` falls into the "no L3 present" branch (L2 <= 8 MB) and gives the 4 MB `NURSERY_SIZE_UNKNOWN_CACHE` fallback rather than half of the real performance-cluster L2.

  https://developer.apple.com/documentation/kernel/1387446-sysctlbyname/determining_system_capabilities#3901385

`hw.l3cachesize` (and Apple's documented `hw.perflevel0.l3cachesize`) both appear unpopulated on my M4 MacBook Air, so this commit doesn't try to recover an "L3" contribution there; heavily-allocating workloads may still want to set `PYPY_GC_NURSERY` higher, but the default is at least closer to correct.

On [rpylean](https://github.com/Julian/rpylean) this definitely looks like a performance improvement, 10% on workloads with 8MB vs 4MB, and closer to 40% with manually-set 64MB nursery -- I'm hoping it'll be obvious to y'all whether this change is safe to make for PyPy itself too.

Refs: 39b141959b (raise default gc nursery size to 4MB)

### DIFF
--- a/rpython/memory/gc/env.py
+++ b/rpython/memory/gc/env.py
@@ -416,7 +416,24 @@ def get_L2cache_darwin():
     on the machine we are running on.
     """
     debug_start("gc-hardware")
-    L2cache = get_darwin_sysctl_signed("hw.l2cachesize")
+    # On Apple Silicon, `hw.l2cachesize` returns the *efficiency* cluster's
+    # L2 (e.g. 4 MB on an M-series chip) rather than the performance
+    # cluster's (16 MB and up). Apple documents per-cluster topology under
+    # `hw.perflevelN.*` and states that "lower values of N indicate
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @Julian

mattip:
Whoopsie - em-dash in comments is not  pypy-friendly. I fixed it in 9b2a31d4c5c

PR #5475, opened 2026-05-17
Add types and coments to determenistic

Neppord:
I wanted to start working on the parseing library again, and thought that adding short descriptions and type anotations would help ai and tools understand the code better.

I have used Qwen3.6 to help me understand and coment the code. I have read every line and reviewd it both by hand and using Qwen multiple times.

I have more documentation in the pipeline, but i did not want to overwhelm the project and also wanted to be able to use the feedback I get in this pr to improve the other documentation.

### DIFF
--- a/rpython/rlib/parsing/deterministic.py
+++ b/rpython/rlib/parsing/deterministic.py
@@ -395,6 +395,7 @@ return ~i""")
         return DFARunner(self)
 
     def make_nondeterministic(self):
+        # type: () -> NFA
         """Convert this DFA into an equivalent NFA."""
         result = NFA()
         result.num_states = self.num_states
@@ -450,6 +451,7 @@ class DFARunner(object):
         self.state = 0
 
     def nextstate(self, char):
+        # type: (str) -> int
         """Advance the DFA by one character. Raises KeyError on invalid transition."""
         self.state = self.automaton[self.state, char]
         return self.state
--- a/rpython/rlib/parsing/deterministic.py
+++ b/rpython/rlib/parsing/deterministic.py
@@ -33,6 +33,7 @@ def compress_char_set(chars):
     return real_result
 
 def make_nice_charset_repr(chars):
+    """Create a human-readable representation of a character set."""
     # Compress the letters & digits
     letters = set(chars) & set("0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
     therest = set(chars) - letters - set(['-'])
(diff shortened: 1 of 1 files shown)


rpython/rlib/parsing/deterministic.py
```diff
@@ -94,6 +100,8 @@ def __repr__(self):
             self.unmergeable_states, self.names)), )
 
     def add_state(self, name=None, final=False, unmergeable=False):
+        # type: (Optional[str], Optional[bool], Optional[bool]) -> int
+        """Add a new state to the automaton and return its id."""
```
StanFromIreland, line 104: These won't be recognised as docstings.

Neppord: why?

Neppord: seams to work fine to me
```
pypy -m pydoc rpython.rlib.parsing.deterministic.make_nice_charset_repr | cat
Help on function make_nice_charset_repr in rpython.rlib.parsing.deterministic:

rpython.rlib.parsing.deterministic.make_nice_charset_repr = make_nice_charset_repr(chars)
    Create a human-readable representation of a character set.
```

rpython/rlib/parsing/deterministic.py
```diff
@@ -215,6 +227,8 @@ def optimize(self):
         return True
 
     def make_code(self):
+        # type: () -> Callable[[str], bool]
+        """Generate and compile a Python function that recognizes the DFA's language."""
```
cfbolz: I would be more specific here and write 'RPython'

cfbolz:
I'm fine with this, apart from my above small nit

rpython/rlib/parsing/deterministic.py
StanFromIreland: Ah, right, comments and blank lines are ignored before the AST is built, sorry!

```py
>>> def f():
...     pass
...     """doc"""
...     
>>> f.__doc__
>>> 
```

Is what I was worried about.

mattip:
For your goal, does this level of annotation do the job? Do you find it helpful? Do the tools work more smoothly?

Neppord:
> For your goal, does this level of annotation do the job? Do you find it helpful? Do the tools work more smoothly?

To be honest it's mostly speculative. But it would help me as a human and I would then guess that it would help a llm. Especially with the correct documentation reading skill.

The types help pycharm, at least in my experience. I have not tried other tools yet.

PR #5476, opened 2026-05-17
Add __eq__ / __ne__ to `r_dict`s

Julian:
Obviously just for untranslated mode to help with writing tests which compare `r_dict`s.


mattip:
Which tests would this enable simplifying?

Julian:
I already essentially worked around it by instead asserting about wrapper objects so if this doesn't make sense fine to close obviously! But originally it was ones whoch go via https://github.com/Julian/rpylean/blob/1f6d2205d3958484b7de351d314a5e02896b9fab/rpylean/environment.py#L1802

Issue #5477, opened 2026-05-17
json serializer not handling some derived classes (Regression)


jwag956:
The system json serializer isn't properly converting certain derived classes correctly in pypy3.11-v7.3.22. This is a regression from v7.3.21.

```
Python 3.11.15 (dc0121776214, Apr 28 2026, 06:06:36)
[PyPy 7.3.22 with GCC Apple LLVM 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> import json
.... from enum import Enum
.... 
.... class alg(int, Enum):
....   algtest = 5
.... 
.... d = dict(alg=alg.algtest)
.... json.dumps(d)
.... 
'{"alg": alg.algtest}'
```

From v7.3.21:
```

[PyPy 7.3.21 with GCC Apple LLVM 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>>> import json
.... from enum import Enum
.... 
.... class alg(int, Enum):
....   algtest = 5
.... 
.... d = dict(alg=alg.algtest)
.... json.dumps(d)
.... 
'{"alg": 5}'
```

StanFromIreland:
It's a regression caused by https://github.com/pypy/pypy/commit/e48df8bbd76421550089ab667a5217ba814d8246, ironically we had a similar issue with reprs in CPython 3.15 recently. I think I can write a little patch.

PR #5478, opened 2026-05-17
Fix `_pypyjson` encoding of int subclasses

StanFromIreland:
Closes https://github.com/pypy/pypy/issues/5477.


mattip:
Thanks @StanFromIreland

Issue #3397, opened 2021-02-10
Provide a stable ABI for PyPy (like CPython's abi3)


gitlab-importer:
In Heptapod by @ShaneHarvey on Feb 10, 2021, 19:39

It would be great for PyPy to add support for a stable ABI similar to abi3 for CPython. Something where a library maintainer could upload a wheel built with say PyPy 3.6 (7.3.1) and that wheel would be guaranteed to work with future PyPy versions (3.7, 3.8, etc...).

Currently, as I understand it, library maintainers can release PyPy wheels but they are only compatible with a specific PyPy version.

For background, the cryptography project has started using rust in its build process which is fine for CPython users but forces all pypy users to install rust. This feature would make it more feasible for them to ship pypy wheels:
- https://github.com/pyca/cryptography/issues/5696
- https://github.com/pyca/cryptography/issues/5806

mattip:
Also see #2617 which floats the idea of mapping C-API functions to fields in a struct.

mattip:
The good news is that for python 3.12, Py_INCREF and Py_DECREF are now true function calls. The bad news is that the `Py_TYPE`/`Py_SIZE`/`Py_REFCNT` macros directly access fields in `PyObject*`, and PyPy's `ob_pypy_link` sits in the struct in a place that will make all those direct field access macros fail. Other than that, we are pretty close to supporting LIMITED_API functions, at least the ones used in cryptograpy. So to actually be able to load CPython ABI3 c-extensions, we would need to
- implement many more of the trivial CPython LIMITED_API functions, and try to work out a strategy for the rarely used ones we will probably never support, like all the embedding and startup configuration ones
- move `ob_pypy_link` to a place that does not interfere. There is precedent for moving it to _before_ the beginning of the struct (CPython's own PyGC_Head/_PyObject_HEAD_EXTRA), which would need to be carefully handled. 
- rename the LIMITED_API exported symbols to drop the Py -> PyPy name mangling
- teach the extension module loader to load ABI3 shared objects

The transition to py3.12 is a good time to try this out and see how far it can get.

mattip:
I asked AI (claude) to take a look at the top PyPI packages to try to figure out which ones use ABI3 packaging. It turns out the maturin (rust-O3) abi-py3x route is increasingly being used by packages. Here is the analysis:

## Real-world ABI3 demand (top PyPI packages)

Source: the download-ranked package list at
[...]

PR #5488, opened 2026-05-29
Use manylinux_2_28 docker images in the rpython github action

mattip:
This is the basis of moving our build images to manylinux_2_28 to pick up gcc14 instead of gcc5 in the current manylinux_2014 images. As a bonus it allows running RPython tests on github actions, in addition to (maybe in place of...) our buildbot infrastructure.
- a new [pypy-ci](https://github.com/pypy/pypy-ci) repo holds docker images based on manylinux_2_28 which is based on almalinux 8, which provides gcc14.
- This PR is based on work done against mattip/pypy in order not to create noise here. All the rpython tests are run in 5 github jobs, each with 4 cores. The entire workflow finishes the rpython testing in about half an hour. More can be done to reduce the total throughput by using the new file-based (instead of directory based) PYPYCHERRYPICK and PYPYIGNORETEST env variables from #5487 

This enables:
- benchmarker2 runs benchmarks with that image, and is live on speed.pypy.org. Once it stabilizes benchmarker will be decommissioned. The benchmarks run on 3 isolated cores with no turbo boost, so the results are very similar to the older benchmarker machine.
- A new service (runs on other isolated cores on benchmarker2) at https://build-summary.pypy.org which can show both buildbot results and gha results.


mattip:
@cfbolz points out this would require an ABI change, 7.4? 8.0? That provides an opportunity (challenge?) to think about finally providing a stable C ABI for PyPY, like abi3 for CPython. See #3397

mattip:
I updated the pypy version to 8.0.0. I think we can move the buildbots to manylinux_2_28

PR #5489, opened 2026-05-30
OpenSSL 4.x support

stefanor:
I made a quick attempt at OpenSSL 4.x support

I haven't tested this very thoroughly, yet. It was enough to get the module to build under 3.11.

### DIFF
--- a/lib-python/2.7/test/test_ssl.py
+++ b/lib-python/2.7/test/test_ssl.py
@@ -408,11 +408,11 @@ class BasicSocketTests(unittest.TestCase):
         # Some sanity checks follow
         # >= 0.9
         self.assertGreaterEqual(n, 0x900000)
-        # < 4.0
-        self.assertLess(n, 0x40000000)
+        # < 5.0
+        self.assertLess(n, 0x50000000)
         major, minor, fix, patch, status = t
         self.assertGreaterEqual(major, 1)
-        self.assertLess(major, 4)
+        self.assertLess(major, 5)
         self.assertGreaterEqual(minor, 0)
         self.assertLess(minor, 256)
         self.assertGreaterEqual(fix, 0)
@@ -2375,8 +2375,7 @@ else:
             with server, \
                     closing(socket.socket()) as sock, \
                     closing(ssl.wrap_socket(sock,
-                                        certfile=certfile,
-                                        ssl_version=ssl.PROTOCOL_TLSv1)) as s:
+                                        certfile=certfile)) as s:
                 try:
                     # Expect either an SSL error about the server rejecting
                     # the connection, or a low-level connection reset (which
@@ -3189,10 +3188,10 @@ else:
                 (['abc', 'def'], 'abc')
             ]
             for client_protocols, expected in protocol_tests:
-                server_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+                server_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
                 server_context.load_cert_chain(CERTFILE)
                 server_context.set_npn_protocols(server_protocols)
-                client_context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+                client_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
                 client_context.load_cert_chain(CERTFILE)
                 client_context.set_npn_protocols(client_protocols)
                 stats = server_params_test(client_context, server_context,
--- a/lib_pypy/_cffi_ssl/_cffi_src/openssl/asn1.py
+++ b/lib_pypy/_cffi_ssl/_cffi_src/openssl/asn1.py
@@ -13,20 +13,13 @@ typedef int... time_t;
 
 typedef ... ASN1_INTEGER;
 
-struct asn1_string_st {
-    int length;
-    int type;
-    unsigned char *data;
-    long flags;
-};
-
-typedef struct asn1_string_st ASN1_OCTET_STRING;
-typedef struct asn1_string_st ASN1_IA5STRING;
-typedef struct asn1_string_st ASN1_BIT_STRING;
-typedef struct asn1_string_st ASN1_TIME;
+typedef ... ASN1_OCTET_STRING;
+typedef ... ASN1_IA5STRING;
+typedef ... ASN1_BIT_STRING;
+typedef ... ASN1_TIME;
 typedef ... ASN1_OBJECT;
-typedef struct asn1_string_st ASN1_STRING;
-typedef struct asn1_string_st ASN1_UTF8STRING;
+typedef ... ASN1_STRING;
+typedef ... ASN1_UTF8STRING;
 typedef ... ASN1_TYPE;
 typedef ... ASN1_GENERALIZEDTIME;
 typedef ... ASN1_ENUMERATED;
--- a/lib_pypy/_cffi_ssl/_cffi_src/openssl/engine.py
+++ b/lib_pypy/_cffi_ssl/_cffi_src/openssl/engine.py
@@ -5,7 +5,10 @@
 from __future__ import absolute_import, division, print_function
 
 INCLUDES = """
+#include <openssl/configuration.h>
+#ifndef OPENSSL_NO_ENGINE
 #include <openssl/engine.h>
+#endif
 """
 
 TYPES = """
--- a/lib_pypy/_cffi_ssl/_cffi_src/openssl/ocsp.py
+++ b/lib_pypy/_cffi_ssl/_cffi_src/openssl/ocsp.py
@@ -23,7 +23,7 @@ static const long OCSP_RESPID_KEY;
 FUNCTIONS = """
 int OCSP_response_status(OCSP_RESPONSE *);
 OCSP_BASICRESP *OCSP_response_get1_basic(OCSP_RESPONSE *);
-int OCSP_BASICRESP_get_ext_count(OCSP_BASICRESP *);
+const int OCSP_BASICRESP_get_ext_count(OCSP_BASICRESP *);
 const ASN1_OCTET_STRING *OCSP_resp_get0_signature(const OCSP_BASICRESP *);
 const Cryptography_STACK_OF_X509 *OCSP_resp_get0_certs(const OCSP_BASICRESP *);
 const ASN1_GENERALIZEDTIME *OCSP_resp_get0_produced_at(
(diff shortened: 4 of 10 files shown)


stefanor:
Got some of the more egregious stuff fixed. Lots of tests are still failing because the old protocols were removed.

mattip:
Is there a cpython version that supports this? Then we could backport the ssl.py and the tests

StanFromIreland:
CPython 3.15, see https://github.com/python/cpython/issues/148600

mattip:
Hmm. That led me to the discussion of [backporting to CPython3.14](https://github.com/python/cpython/pull/149783#issuecomment-4507558612), where the conclusion was to wait until 3.15.1 for the dust to settle around the version. I think that means around the end of the year. @stefanor is there a demand for this that would drive merging this earlier?

stefanor:
I don't have a timeline for when we'll need this in Debian yet. Obviously a lot of software needs to add support, still. It's at the point now where bugs are being filed to try to fix this situation upstream.

My expectation is that in a couple of months, I'll need to carry a patch like this.

stefanor:
The OpenSSL maintainer in Debian has requested a slot for a transition: https://bugs.debian.org/1140179
(But clearly lots of packages are going to need to be fixed first)

mattip:
From that email chain, I see that
> ... 3.5 (same as in trixie), which is an LTS release supported 
until April 2030.

>This is not a one-off issue:
>OpenSSL plans to have new LTS releases every second year in April,
always in our release years.

So it would be good to have a policy around version support. In any case, PyPy will try to follow whatever CPython does for already released versions, i.e. whatever is decided for CPython3.14 at this point.

loqs:
> Got some of the more egregious stuff fixed. Lots of tests are still failing because the old protocols were removed.

Only SSL3 was removed. TLS 1.0, 1.1 and 1.2 are still present in OpenSSL 4.0 although TLS 1.0 and 1.1 are excluded by the default cipher spec. Most of the failing tests appear to be caused by use of the removed methods in https://github.com/stefanor/pypy/blob/fc2f20bd2152b9155af0a854fbbfef9c24d19e57/lib_pypy/_cffi_ssl/_stdssl/__init__.py#L993-L1002.

The switch to solely using `TLS_method()` could be done first and is backwards compatible?

loqs:
@stefanor with the following additional commit applied I have 5 errors for pypy2 and after rebasing to the py312 branch reduces that to 1 flaky test failure:
```patch
diff --git a/lib_pypy/_cffi_ssl/_cffi_src/openssl/ssl.py b/lib_pypy/_cffi_ssl/_cffi_src/openssl/ssl.py
index 1c8a115520b..a8252d1255f 100644
--- a/lib_pypy/_cffi_ssl/_cffi_src/openssl/ssl.py
+++ b/lib_pypy/_cffi_ssl/_cffi_src/openssl/ssl.py
@@ -633,8 +633,8 @@ static const long Cryptography_HAS_STATUS_REQ_OCSP_RESP = 1;
 static const long Cryptography_HAS_TLSEXT_STATUS_REQ_TYPE = 1;
 static const long Cryptography_HAS_RELEASE_BUFFERS = 1;
 static const long Cryptography_HAS_OP_NO_COMPRESSION = 1;
-static const long Cryptography_HAS_TLSv1_1 = 0;
-static const long Cryptography_HAS_TLSv1_2 = 0;
+static const long Cryptography_HAS_TLSv1_1 = 1;
+static const long Cryptography_HAS_TLSv1_2 = 1;
 static const long Cryptography_HAS_SSL_OP_MSIE_SSLV2_RSA_PADDING = 1;
 static const long Cryptography_HAS_SSL_OP_NO_TICKET = 1;
 static const long Cryptography_HAS_SSL_SET_SSL_CTX = 1;
diff --git a/lib_pypy/_cffi_ssl/_stdssl/__init__.py b/lib_pypy/_cffi_ssl/_stdssl/__init__.py
index bebf298e503..abe119ad52b 100644
--- a/lib_pypy/_cffi_ssl/_stdssl/__init__.py
+++ b/lib_pypy/_cffi_ssl/_stdssl/__init__.py
@@ -103,13 +103,6 @@ SSL_CB_MAXLEN=128
 if lib.Cryptography_HAS_SSL2:
     PROTOCOL_SSLv2  = 0
 SSLv3_method_ok = False
-if lib.Cryptography_HAS_SSL3_METHOD:
-    # Some Ubuntu systems disable SSLv3
-    ctx = lib.SSL_CTX_new(lib.SSLv3_method())
-    if ctx:
-        PROTOCOL_SSLv3  = 1
-        lib.SSL_CTX_free(ctx)
-        SSLv3_method_ok = True
         
 PROTOCOL_SSLv23 = 2
 PROTOCOL_TLS    = PROTOCOL_SSLv23
@@ -120,7 +113,7 @@ if lib.Cryptography_HAS_TLSv1_2:
     PROTOCOL_TLSv1_2 = 5
 HAS_SSLv2 = bool(lib.Cryptography_HAS_SSL2)
 HAS_SSLv3 = SSLv3_method_ok
-HAS_TLSv1 = False # XXX
+HAS_TLSv1 = True # XXX
 HAS_TLSv1_1 = bool(lib.Cryptography_HAS_TLSv1_1)
 HAS_TLSv1_2 = bool(lib.Cryptography_HAS_TLSv1_2)
 HAS_TLSv1_3 = bool(lib.Cryptography_HAS_TLSv1_3)
@@ -990,18 +983,18 @@ class _SSLContext(object):
     def __new__(cls, protocol):
         self = object.__new__(cls)
         self.ctx = ffi.NULL
-        if HAS_TLSv1 and protocol == PROTOCOL_TLSv1:
-            method = lib.TLSv1_method()
+        if protocol == PROTOCOL_TLSv1:
+            method = lib.TLS_method()
+            self_min = self_max = PROTO_TLSv1
         elif lib.Cryptography_HAS_TLSv1_1 and protocol == PROTOCOL_TLSv1_1:
-            method = lib.TLSv1_1_method()
+            method = lib.TLS_method()
+            self_min = self_max = PROTO_TLSv1_1
         elif lib.Cryptography_HAS_TLSv1_2 and protocol == PROTOCOL_TLSv1_2 :
-            method = lib.TLSv1_2_method()
-        elif SSLv3_method_ok and protocol == PROTOCOL_SSLv3:
-            method = lib.SSLv3_method()
-        elif lib.Cryptography_HAS_SSL2 and protocol == PROTOCOL_SSLv2:
-            method = lib.SSLv2_method()
+            method = lib.TLS_method()
+            self_min = self_max = PROTO_TLSv1_2
         elif protocol == PROTOCOL_SSLv23:
             method = lib.TLS_method()
+            self_min = self_max = 0
         else:
             raise ValueError("invalid protocol version")
 
@@ -1009,6 +1002,12 @@ class _SSLContext(object):
         __pypy__.add_memory_pressure(1000)
         if self.ctx == ffi.NULL:
             raise ssl_error("failed to allocate SSL context")
+        if self_min:
+            if not lib.SSL_CTX_set_min_proto_version(self.ctx, self_min):
+                raise ValueError("invalid or unsupported protocol version")
+        if self_max:
+            if not lib.SSL_CTX_set_max_proto_version(self.ctx, self_max):
+                raise ValueError("invalid or unsupported protocol version")
         self._post_handshake_auth = 0;
         self._protocol = protocol
         self.hostflags = lib.X509_CHECK_FLAG_NO_PARTIAL_WILDCARDS
@@ -1020,10 +1019,8 @@ class _SSLContext(object):
 
         # Defaults
         options = lib.SSL_OP_ALL & ~lib.SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS
-        if not lib.Cryptography_HAS_SSL2 or protocol != PROTOCOL_SSLv2:
-            options |= lib.SSL_OP_NO_SSLv2
-        if not SSLv3_method_ok or protocol != PROTOCOL_SSLv3:
-            options |= lib.SSL_OP_NO_SSLv3
+        options |= lib.SSL_OP_NO_SSLv2
+        options |= lib.SSL_OP_NO_SSLv3
         # Minimal security flags for server and client side context.
         # Client sockets ignore server-side parameters.
         options |= lib.SSL_OP_NO_COMPRESSION
@@ -1037,11 +1034,7 @@ class _SSLContext(object):
 
         # A bare minimum cipher list without completely broken cipher suites.
         # It's far from perfect but gives users a better head start.
-        if lib.Cryptography_HAS_SSL2 and protocol == PROTOCOL_SSLv2:
-            # SSLv2 needs MD5
-            default_ciphers = b"HIGH:!aNULL:!eNULL"
-        else:
-            default_ciphers = b"HIGH:!aNULL:!eNULL:!MD5"
+        default_ciphers = b"HIGH:!aNULL:!eNULL:!MD5"
         if not lib.SSL_CTX_set_cipher_list(self.ctx, default_ciphers):
             lib.ERR_clear_error()
             raise SSLError("No cipher can be selected.")
```

stefanor:
~~@loqs: That doesn't look right to me. If you use `TLS_method` to make what you expect to be a TLS 1.0 connection, how is OpenSSL supposed to know that it's a TLS 1.0 connection? Also, as you say, TLS 1.0 is disabled by default. So... you'd need to do something to enable it.~~

stefanor:
Clearly I missed the important bit when I skimmed this earlier. Yeah, that does make sense.

stefanor:
@loqs: Not that your approach does not match cPython's. Rather than emulating the old methods, they simply remove support for them (as my branch is doing). That said, they haven't backported this to < 3.15 yet, so we'll see where that goes...

```
# Python 3.16.0a0
>>> import ssl
>>> ssl.HAS_TLSv1
False
>>> ssl.HAS_TLSv1_1
False
>>> ssl.HAS_TLSv1_2
False
>>> ssl.HAS_TLSv1_3
True
```

loqs:
> @loqs: Not that your approach does not match cPython's. Rather than emulating the old methods, they simply remove support for them (as my branch is doing). That said, they haven't backported this to < 3.15 yet, so we'll see where that goes...

I missed that. Your approach works perfectly rebased onto py13 https://github.com/loqs/pypy/tree/py3.12-openssl4. https://github.com/loqs/pypy/tree/py2.7-openssl4 is my attempt at addressing the remaining test failures by matching the 3.12 branch/cpython.

Edit:

py3.12 4 failures as it is pulling cpythons 3.12's ssl tests not 3.15's. py2.7 no test failures.

loqs:
> @loqs: Not that your approach does not match cPython's.

I missed that. Matching upstream cpython changes instead (I included the upstream commit each change is based on):
- https://github.com/loqs/pypy/commits/py2.7-openssl4/
- https://github.com/loqs/pypy/tree/py3.12-openssl4

The first commit on py2.7-openssl4 has the greatest divergence from cpython due to the version gap between 2.7 and 3.15.

Issue #5490, opened 2026-05-31
Weird regression on 7.3.23 with finally blocks catching their own errors


cobaltt7:
<https://github.com/psf/black>'s CI recently [started failing](https://github.com/psf/black/actions/runs/26697388270/job/78684120782?pr=5159#step:5:158) and I eventually narrowed it down to a regression between PyPy 7.3.22 and 7.3.23.
It appears that `finally` blocks are catching errors raised in themselves, but only if a previous `except` block ran a `return` statement? (In the MRE below, it does not run twice if I remove either `raise Exception`, or if I change the `return` to a `pass`, or if I move the `return` into the `try`)
Something's super weird here and I'm not totally sure what
### Minimal Reproducible Example
```py
def foo():
    try:
        raise Exception
    except:
        return
    finally:
        print("running finally block")
        raise Exception


foo()
```

### 7.3.22 output (expected; matches CPython)
```py
running finally block
Traceback (most recent call last):
  File "...\test.py", line 11, in <module>
    foo()
  File "...\test.py", line 8, in foo
    raise Exception
Exception
```

### 7.3.23 output (unexpected)
```py
running finally block
running finally block
Traceback (most recent call last):
  File "...\test.py", line 8, in foo
    raise Exception
Exception

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "...\test.py", line 11, in <module>
    foo()
  File "...\test.py", line 8, in foo
    raise Exception
Exception
```

mattip:
Bahh. There were no tests for this. We recently revise exception handling to be more like CPython (using an exceptiontable) and this case got missed. Should be fixed in #5491.

cobaltt7:
Thanks! Any ETA on a new release?

mattip:
Not really. There are a lot of balls in the air: moving to manylinux2_28 builds, prototyping using abi3 wheels in pypy, pushing 3.12 forward. I think it will be a couple of months at least until the next release.

PR #5491, opened 2026-05-31
fix unwind_fblock(F_HANDLER_CLEANUP) for bare except (issue 5490)

mattip:
During testing, I added dis.dis to show the exceptiontable, which made the problem very evident.

In `unwind_fblock(F_HANDLER_CLEANUP)` (the code path taken when return/break/continue exits an except handler), always emit __two__ `_POP_BLOCK` pseudo-ops before `POP_EXCEPT` — one to close the inner cleanup scope (`SETUP_CLEANUP` -> `cleanup_end`) and one to close the outer cleanup scope (`SETUP_CLEANUP` -> `outer_cleanup`). Previously, bare `except:` got zero, named `except E as x:` got one. The normal exit path at lines 1027–1029 already emitted two; this unwind path now matches it.

With both scopes closed, `POP_EXCEPT` falls under the `SETUP_FINALLY` scope (depth 0, no `lasti`), and the inlined `finally` body has no exception table coverage at all.

Fixes issue #5490

### DIFF
--- a/pypy/interpreter/astcompiler/codegen.py
+++ b/pypy/interpreter/astcompiler/codegen.py
@@ -346,8 +346,8 @@ class PythonCodeGenerator(assemble.PythonCodeMaker):
             self.emit_op(ops.POP_TOP)
             self.no_position_info()
         elif kind == F_HANDLER_CLEANUP:
-            if fblock.datum:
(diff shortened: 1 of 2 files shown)


StanFromIreland (approved):
LGTM

Issue #5340, opened 2025-10-12
missing globals from `socket` module compared to CPython 3.11


eqvinox:
Just tried running some network software on `pypy`; noticed there are a bunch of module level constants missing from the `socket` module (the one I ran into is `SO_PEERCRED`).

Arguably some of these are a bit exotic (e.g. `CAN_*` and `TIPC_*`) but the `SO*` `IP*` `TCP*` `MSG*` and `SCM_CREDENTIALS` are kinda important for any nontrivial networking code.

The actual delta to CPython 3.11 (on Linux) is:
```
AF_ALG
AF_CAN
AF_DECnet
AF_QIPCRTR
AF_RDS
AF_TIPC
AF_VSOCK
ALG_OP_DECRYPT
ALG_OP_ENCRYPT
ALG_OP_SIGN
ALG_OP_VERIFY
ALG_SET_AEAD_ASSOCLEN
ALG_SET_AEAD_AUTHSIZE
ALG_SET_IV
ALG_SET_KEY
ALG_SET_OP
ALG_SET_PUBKEY
BTPROTO_HCI
BTPROTO_L2CAP
BTPROTO_RFCOMM
BTPROTO_SCO
CAN_BCM
CAN_BCM_CAN_FD_FRAME
CAN_BCM_RX_ANNOUNCE_RESUME
CAN_BCM_RX_CHANGED
CAN_BCM_RX_CHECK_DLC
CAN_BCM_RX_DELETE
CAN_BCM_RX_FILTER_ID
CAN_BCM_RX_NO_AUTOTIMER
CAN_BCM_RX_READ
CAN_BCM_RX_RTR_FRAME
CAN_BCM_RX_SETUP
CAN_BCM_RX_STATUS
CAN_BCM_RX_TIMEOUT
CAN_BCM_SETTIMER
CAN_BCM_STARTTIMER
CAN_BCM_TX_ANNOUNCE
CAN_BCM_TX_COUNTEVT
CAN_BCM_TX_CP_CAN_ID
CAN_BCM_TX_DELETE
CAN_BCM_TX_EXPIRED
CAN_BCM_TX_READ
CAN_BCM_TX_RESET_MULTI_IDX
CAN_BCM_TX_SEND
CAN_BCM_TX_SETUP
CAN_BCM_TX_STATUS
CAN_EFF_FLAG
CAN_EFF_MASK
CAN_ERR_FLAG
CAN_ERR_MASK
CAN_ISOTP
CAN_J1939
[...]

PR #5492, opened 2026-06-01
add missing socket constants (issue 5340)

mattip:
To add more global constants to the python3 _socket module, first we need to add them to rsocket.


mattip:
Closes #5340. As long as I am touching rsocket, add the implementation of `dup(fd, inherit=False)` for windows

PR #5501, opened 2026-06-09
fix buffer-related pypyjit tests

mattip:
Fallout from #5444. The jit traces are slightly different now
- when exporting a buffer we increment its exports count
- when releasing a buffer we decrement its exports count
- a loop counter moved


cfbolz:
lgtm, thx!

PR #5502, opened 2026-06-10
cache hash of tuples

mattip:
CPython started caching hashes of tuples in python/cpython#131529, and quoting from the PR 
> in a recent experiment to add a tuple hash cache back in, the mdp benchmark increased by 86%.

Admittedly, there was no measurable improvement on any other benchmark, but it also seems to have no downside, including for memory usage when measured with max_rss.

Like them, I noticed that tuple hashing featured heavily in the bm_mdp benchmark. This change made that benchmark 2.7x faster with `--jit off` and 3.7x faster with `jit on`, mirroring CPython's results.

Additionally, the CPython PR carefully resets the `ob_hash` to -1 when reusing tuple objects, but PyPy does not (yet?) reuse tuples so the change is much simpler


cfbolz:
let's add a test that the caching works? eg by having a tuple entry where the `__hash__` has a side effect that we can check in the test.

mattip:
Test added in 81dc8c8, after refactoring bare tests into apptests.

pypy/objspace/std/test/apptest_specializedtuple.py
```diff
@@ -0,0 +1,4 @@
+# spaceconfig = {"objspace.std.withspecialisedtuple": true}
+import sys, os
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from apptest_tuple import *
```
mattip, line 4: This replaces the test_specialisedtupleobject AppTestAll

pypy/objspace/std/test/test_specialisedtupleobject.py
```diff
@@ -71,6 +70,16 @@ def hash_test(values, must_be_specialized=True):
...
+    def test_hash_consistency(self):
+        # make sure that the two copies of the hash implementation are the same
+        w = self.space.wrap
+        w_tuple1 = W_TupleObject([w(5), w(3), w(99)])
+        w_tuple2 = W_TupleObject([w(5), w(3), w(99)])
+        w_tuple3 = W_TupleObject([w(5), w(3), w(99), w(-1)])
+        w_tuple4 = W_TupleObject([w(5), w(3), w(9), w(-1)])
+        for w_tup in w_tuple1, w_tuple2, w_tuple3, w_tuple4:
+            assert w_tup._descr_hash_unroll(self.space) == w_tup._descr_hash_jitdriver(self.space)
+
```
mattip, line 82: This was the only test in test_tuple.py that could not be converted into an apptest

pypy/objspace/std/test/apptest_tuple.py
```diff
@@ -0,0 +1,307 @@
...
+    u = subclass_with_new([1, 2], newarg=3)
+    assert u.newarg == 3
+
+    class subclass_with_init(tuple):
+        def __init__(self, arg, newarg=None):
+            self.newarg = newarg
+    u = subclass_with_init([1, 2], newarg=3)
+    assert u.newarg == 3
+
+def test_hash_cache():
```
mattip, line 296: Added from review

cfbolz:
looks good, I'm merging

PR #5519, opened 2026-07-03
gh-5518: make dict.update consume an iterable argument lazily

### LINKED ISSUE #5518: dict.update() updates lazily from a generator
I have the following code:
```python
exception_subtypes = {
    'BaseException' : [
        'BaseExceptionGroup',
        'Exception',
        'GeneratorExit',
        'KeyboardInterrupt',
        'SystemExit',
    ],
    'BaseExceptionGroup' : [
        'ExceptionGroup',
    ],
    # ...
}

exception_supertypes = {'BaseException': ['BaseException']}
exception_supertypes.update(
    # Rebuild the exception type MRO, knowing that the 'exception_subtypes' dict
    # was built breadth-first, listing all parents before their children.
    (child, [child] + exception_supertypes[parent])
    for parent, children in exception_subtypes.items()
    for child in children
)
```
In CPython, this allows for an efficient linear-time supertypes dict creation, by inserting one item from the generator expression at a time before requesting the next item. It fails in PyPy, apparently because the first items have not yet been inserted into the supertypes dict (or can at least not be looked up there) when it reaches the `BaseExceptionGroup` entry in the subtypes dict that the generator iterates over.

The error is:
```
    (child, [child] + exception_supertypes[parent])
                      ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'BaseExceptionGroup'
```

It can certainly be argued whether there _should_ be a guarantee regarding the time of insertions, but I don't think it's wrong to assume that once an item has been retrieved from the generator (and iteration continues), it should be stored in the dict being updated – because where else should it be stored? The usage of a generator (expression), as opposed to passing a prepared list of items into `.update()`, is to apply the updates as they are being generated, and at least not using intermediate storage for them.

cfbolz:
fixes #5518

### DIFF
--- a/pypy/objspace/std/dictmultiobject.py
+++ b/pypy/objspace/std/dictmultiobject.py
@@ -1416,14 +1416,10 @@ def update1(space, w_dict, w_data):
     w_method = space.findattr(w_data, space.newtext("keys"))
     if w_method is None:
         # no 'keys' method, so we assume it is a sequence of pairs
-        data_w = space.listview(w_data)
-        if data_w:
-            update1_pairs(space, w_dict, data_w)
+        update1_pairs(space, w_dict, w_data)
     else:
         # general case -- "for k in o.keys(): dict.__setitem__(d, k, o[k])"
-        data_w = space.listview(space.call_function(w_method))
-        if data_w:
-            update1_keys(space, w_dict, w_data, data_w)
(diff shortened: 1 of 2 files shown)


mattip:
@cfbolz can we merge this?

PR #5520, opened 2026-07-04
change getrecursionlimit, setrecursionlimit to mean python stack frames

mattip:
Until now the `sys.setrecursionlimit` and `sys.getrecursionlimit` interfaces on CPython relate to python frame depth, but on PyPy related to the c stack depth. In 238c2946b2c, I change the C stack overflow check to actually get the stack size at startup and check that the interpreter is not running out of stack. This check happens on almost every function call. This PR extends that work to check the pytohn frame depth for CPython compatibility. It means we have a new execution context attribute `recursion_remaining`, and check on entering every frame that the value is valid. By setting it high and decrementing, we do not need to look up the limit each time, we only decrement and check `>=0`. When exiting a frame we increment the value.

This does add some overhead, I want to benchmark to see how much.

### DIFF
--- a/pypy/interpreter/executioncontext.py
+++ b/pypy/interpreter/executioncontext.py
@@ -33,7 +33,8 @@ class ExecutionContext(object):
     def __init__(self, space):
         self.space = space
         self.topframeref = jit.vref_None
-        # Python-level recursion counter, counts down like CPython's py_recursion_remaining
+        # Python-level recursion counter, counts down
+        # In CPython it is called py_recursion_remaining
         self.recursion_remaining = space.sys.recursionlimit
         # this is exposed to app-level as 'sys.exc_info()'.  At any point in
         # time it is the exception caught by the topmost 'except ... as e:'
@@ -87,16 +88,21 @@ class ExecutionContext(object):
     def enter(self, frame):
         if self.space.reverse_debugging:
             self._revdb_enter(frame)
-        remaining = self.recursion_remaining
-        if remaining <= 0:
-            raise oefmt(self.space.w_RecursionError,
-                        "maximum recursion depth exceeded")
-        self.recursion_remaining = remaining - 1
+        # recursion_checking is a quasi-immutable flag: if noone called
+        # sys.setrecursionlimit() it stays False and theJIT folds this
+        # block away, the per-frame counter costs nothing on the hot path.
+        if self.space.actionflag.recursion_checking:
+            remaining = self.recursion_remaining
+            if remaining <= 0:
+                raise oefmt(self.space.w_RecursionError,
+                            "maximum recursion depth exceeded")
+            self.recursion_remaining = remaining - 1
         frame.f_backref = self.topframeref
         self.topframeref = jit.virtual_ref(frame)
 
     def leave(self, frame, w_exitvalue, got_exception):
-        self.recursion_remaining += 1
+        if self.space.actionflag.recursion_checking:
+            self.recursion_remaining += 1
         try:
             if self.profilefunc:
                 self._trace(frame, 'leaveframe', w_exitvalue)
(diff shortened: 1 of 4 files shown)


mattip:
This has a real effect on the go benchmark. It is deeply recursive, so the jit cannot cleanly combine and eliminate the inc/dec of the counter. I will try to work out a more JIT freindly version.

mattip:
The last commit hides the python stack count check behind a gate. If `sys.setrecursionlimit` has never been set, the gate is False, the JIT sees it as constant, and the check is optimized away. Once the function is called (usually for tests), the gate is True and the interpreter and the JIT will more carefully check the limit on every frame boundary.

cfbolz:
is this really worth it, just to pass some extra cpython tests? I'd rather annotate them as cpython_only upstream. Also, IIRC the details around that completely change again in 3.13 or 3.14 (which to my mind shows that even CPython doesn't consider the details important).

mattip:
Hmm. What do you think about declaring `sys.setrecursionlimit()` CPython only and removing the function entirely? That way tests will crash rather than hang

cfbolz:
no, sometimes the API is really useful to set the recursionlimit to something really high. which tests are hanging? should I do a PR to CPython about them, to mark them as `@cpython_only`?

mattip:
If we are going to keep the API, I think this PR provides a nice way to support it. No cost to those who don’t use it, exact CPython 3.12 semantics to those who do. The [last commit benchmark comparison](https://speed.pypy.org/comparison/?exe=24:L:py3.12,24:L:recursion&ben=all&env=4&hor=true&bas=24:L:py3.12&chart=normal+bars) shows parity with py3.12 when `sys.recursion` limit is not set.

BarrensZeppelin:
Setting the limit higher with `sys.setrecursionlimit` is required in some cases with deep recursion (not for CPython compatibility). Wouldn't the change negatively affect performance in these cases?

mattip:
PyPy has never really respected that logic. There was some kind of limit, but no 1:1 correspondence with python stack frames. There is a C stack check when entering every C function that there is still enough room for another C frame. The two were conflated, when calling sys.setrecursionlimit, it would change the C stack check limit. 

> Setting the limit higher with sys.setrecursionlimit is required in some cases with deep recursion

Yes, after this PR that will mean a performance hit, since now we have to start checking the python stack count. The C limit was clamped to 75% of stack in 238c2946b2c. Anything more than that raises a python RecursionError, regardless of the value of `sys.setrecursionlimit`

BarrensZeppelin:
I've used `sys.setrecursionlimit(1<<30)` a lot for "unbounded recursion". Would that immediately raise now?

mattip:
Not immediately, no, why would it? It sets a limit. At certain times as the interpreter runs, it checks X against that limit. On PyPy, X is some factor multiplied by the C stack depth. On CPython, X is the true python stack count. In both CPython and PyPy there is a C-level check to prevent stack overflow. After 238c294, that check on PyPy is more exact, like CPython, so where before that commit you might have gotten a stack overflow segfault, now you should get a clean RecursionError. But in both cases, setting the limit to (effectively) infinite makes the python stack check useless, and depends on the C check to keep the stack from overflowing.

What this PR changes is that now when you set a limit, there are two X checks: the previous one for the C stack and the one in this PR for the python stack. So if you set it low, like 200, PyPy will faithfully try to raise a RecursionError when the python stack hits 201. Also, when a test probes its current recursion depth and then sets the limit a couple frames above, PyPy will, at least, not hang. I think @cfbolz is arguing we should just skip those tests and keep the same 'X approximate frames' check we have before this PR. 

Truthfully, I am beginning to come around to that viewpoint too. This PR is just to hard to explain. But if we reject this, we should document the 'approximate frames' difference.

BarrensZeppelin:
> Not immediately, no, why would it? It sets a limit. 

Great. I read the message for https://github.com/pypy/pypy/commit/238c2946b2c6818905fd69511ee91ee87c811b34 as imposing a limit on the value passed to `sys.setrecursionlimit`.

mattip:
A new test in 3.12 tripped across this and caused a buildbot worker hang. I skipped the test in 3ce98a3aa42. So far I can't see any real world need for this though, so we will keep skipping tests that hit it and hang.

PR #5522, opened 2026-07-07
update pypy version to 8.0.0

mattip:
Update the pypy version, as part of the change to buildbot to use maylinux_2_28 images. See #5488. This will put a burden on the small number of library packagers who still stick with pypy (cryptology for instance) so it is worth some dicusssion.

In Cpython 3.12 the limited api calls to `Py_DECREF` and `Py_INCREF` were finally turned into function calls, I wonder how close we are to being able to support ABI3 c-extensions in first release of py3.12, which would make packages like cryptology very happy. So this is also related to #5513 to examine the function signature and macro compatibility.


mattip:
I guess we can state in the release notes: "we usually recommend users update, but if you use c-extensions you may want to stick with 7.3 until maintainers can release 8.0 packages".

PR #5531, opened 2026-07-22
Implementation Remote Debugger for Windows

philipplc:
- Added Windows support for PyPy’s remote debugger.
- Located pypysig via PE-section parsing with mmap.
- Used Windows process-memory APIs for remote script injection.
- Preserved Linux support and handle Win64 struct offsets.


cfbolz:
it would be great if you could get some of the unit tests in extra_tests for this to run on windows

cfbolz:
merged, thanks for your work @philipplc!

PR #5533, opened 2026-07-24
implement PEP709 inline comprehensions

### LINKED ISSUE #5496: Py 3.12 PEP 709 comprehension inlining
CPython 3.12 inlines list/set/dict comprehensions into the enclosing frame instead of creating separate code objects. PyPy and python3.11 create separate code objects (`_compile_comprehension` -> `sub_scope`  separate `CALL_FUNCTION`).                                                                                                

Observable effects the stdlib tests for: 
- symtable no longer produces child symbol tables for comprehensions
- locals() inside a comprehension sees outer scope variables
- No separate frame in tracebacks
- New `LOAD_FAST_AND_CLEAR` opcode needed                                                                                                                                                     
                                                                                                                                                                                              Options: implement it, or keep old semantics and patch the 3.12 `symtable.py` - but dis and frame inspection tests will likely still fail.

mattip:
Fixes #5496 by implementing inline comprehensions. Along with the new `_inline_comprehension_children` and `_inline_comprehension_children_module`, there is a new opcode `LOAD_FAST_AND_CLEAR` and a `fasthidden` dict. Still to do, among others: python3.11 has a MAKE_CELL opcode that PyPy never implemented, it is needed to create a new cell per execution of comprehension.

### DIFF
--- a/pypy/interpreter/astcompiler/test/apptest_compiler.py
+++ b/pypy/interpreter/astcompiler/test/apptest_compiler.py
@@ -138,3 +138,25 @@ def test_backward_jump_has_lineno():
                if ins.opname == 'JUMP_ABSOLUTE']
     assert len(linenos) > 0
     assert all(l is not None for l in linenos)
+
+
+def test_inlined_comprehension_private_name_reuse():
+    x = 3
+
+    def f():
+        [x for x in [1]]
+        return [x for _ in [1]]
+
+    assert f() == [3]
+
+
+def test_inlined_comprehension_iterator_position():
+    import dis
+
+    code = compile("[x for x in BrokenIter()]", "<test>", "exec")
+    get_iter = [ins for ins in dis.get_instructions(code)
+                if ins.opname == "GET_ITER"][0]
+    position = get_iter.positions
+    assert (position.lineno, position.end_lineno,
+            position.col_offset, position.end_col_offset) == \
+           (1, 1, 12, 24)
--- a/pypy/interpreter/astcompiler/test/apptest_compiler.py
+++ b/pypy/interpreter/astcompiler/test/apptest_compiler.py
@@ -102,3 +102,39 @@ def test_compile_ast_object_pep695_type_alias():
     for src in ("type X = int\n", "type Stack[T] = list[T]\n"):
         tree = compile(src, "<test>", "exec", PyCF_ONLY_AST)
         compile(tree, "<test>", "exec")
+
+
+def test_extended_arg_keeps_source_position():
+    # An instruction needing an EXTENDED_ARG prefix (here UNPACK_EX for
+    # 'a, *b, c', oparg (1<<8)|1 == 257) must keep its source position on the
+    # real opcode, not only on the EXTENDED_ARG.
+    import dis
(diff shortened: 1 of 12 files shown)


mattip:
This seems to work. I admit I let claude AI code this and only monitored what it changed.

cfbolz:
there are still a bunch of bugs in this. it doesn't follow cpython's approach enough, imo. I pushed two tests that fail, and also test_dictcomps, test_listcomps, test_setcomps have a bunch of failures.

mattip:
Thanks I will get back to this in two weeks or so

pypy/interpreter/astcompiler/test/apptest_compiler.py
```diff
@@ -102,3 +102,61 @@ def test_compile_ast_object_pep695_type_alias():
...
+
+
+def test_inlined_comprehension_private_name_reuse():
+    x = 3
+
+    def f():
+        [x for x in [1]]
+        return [x for _ in [1]]
+
+    assert f() == [3]
```
youknowone, line 150: This test fails with CPython 3.14
```
  File "x.py", line 10, in <module>
    test_inlined_comprehension_private_name_reuse()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "x.py", line 8, in test_inlined_comprehension_private_name_reuse
    assert f() == [3]
           ~^^
  File "x.py", line 6, in f
    return [x for _ in [1]]
            ^
UnboundLocalError: cannot access local variable 'x' where it is not associated with a value
```

mattip: We have quite a while till we support 3.14 syntax

youknowone: The test itself is looking like illegal python code.
If that fails on 3.14, hardly it can be legal on 3.12 I guess. (sorry, not directly tested it on 3.12)

youknowone:
I opened a draft PR #5562 that solves more cases. I will not convert it to open PR without any guidance because it introduced localsplus, which is 3.12 spec but not implemented in PyPy yet. I am not sure which way is your intention about localsplus yet.
If you leave guidance to resolve it in preferred design of PyPy, I can try more. Otherwise please cherry-pick any useful hunks or ignore or close it. I don't mind either way.

pypy/interpreter/astcompiler/test/apptest_compiler.py
cfbolz: tbh, if this test fails it looks like a pretty clear bug in CPython? it works in CPython 3.11, and fails in 3.12. I reported it here: https://github.com/python/cpython/issues/156664

cfbolz:
we should note that the cpython devs are pretty unhappy with their own implementation, see this issue: https://github.com/python/cpython/issues/124697

mattip:
Removing this from the next milestone, since it is a (mostly) non-user-facing feature. We could almost call it an implementation detail. Let's wait to see what CPython does about the edge cases and whether they will refactor it.

PR #5535, opened 2026-07-27
count an empty needle in the code point domain

youknowone:
descr_count passed the byte bounds from _unwrap_and_compute_idx_params straight to rstring's search, which answers end - start + 1 for an empty needle, so '\xe9\xe8\xe9\xe8\xe9'.count('') was 11 rather than 6. Answer it from _codepoints_in_utf8 instead, with start_index > end_index giving 0 as rstring.py:827 does, and the whole-string case reading self._len() so the call stays O(1) -- rutf8.codepoints_in_utf8 is a linear scan.

test_count_unicode in apptest_unicode.py now passes; test_unicode, test_bytes, test_string and test_userstring stay green.

### DIFF
--- a/pypy/objspace/std/test/test_unicodeobject.py
+++ b/pypy/objspace/std/test/test_unicodeobject.py
@@ -1001,6 +1001,18 @@ class AppTestUnicodeString:
         assert u'aaa'.count(u'a', 0, -10) == 0
         assert u'ababa'.count(u'aba') == 1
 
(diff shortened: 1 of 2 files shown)


cfbolz:
wow, that's an embarrassing bug. however, one of the unit tests is still failing, please fix it, then I'll review.

youknowone:
I am a little bit surprised this was not covered by CPython unit tests

cfbolz:
thank you!

PR #5537, opened 2026-07-30
jit: fix get_fielddescr_index_in() double-counting the fields that precede an inlined sub-struct

youknowone:
The bug doesn't affect actual code because pypy doesn't hit this line.

The description below and test is AI-generated.

----

The three conditions

The wrong index is produced only when all three hold at once. Each one on its own is harmless, which is why the defect sat in the tree undetected.

1. The struct must contain an inlined sub-struct.
The bad line lives in the elif isinstance(FIELD, lltype.Struct) branch. A struct whose fields are all primitives or pointers never enters it, and the function is then a plain 0,1,2,… counter that is correct.

2. That inlined sub-struct must be preceded by at least one already-counted field.
The recursion is seeded with the running cur_index, so its not-found answer is a running total, not an increment — but the caller adds it with +=. When the inlined struct sits at position 0 the running total is 0, and adding it twice adds nothing, so the answer comes out right by accident:
```
S2(inner, c)        cur_index was 0, recursion reports total=2
  buggy : cur_index += 2  ->  2
  fixed : cur_index  = 2  ->  2      AGREE

S(a, inner, c)      cur_index was 1, recursion reports total=3
  buggy : cur_index += 3  ->  4
  fixed : cur_index  = 3  ->  3      DISAGREE
```
3. The field being looked up must lie after the inlined sub-struct — not before it and not inside it.
Fields before it return before the bad line ever runs, and fields inside it are returned straight out of the recursion (if r >= 0: return r), which was seeded correctly. Only what follows the inlined struct inherits the corrupted counter, together with the not-found total that an enclosing level would consume:

S(a, inner(x, y), c, d)      flattened: a=0 x=1 y=2 c=3 d=4

```
field  where                       expected  answer
a      before the inlined struct   0         0
x      inside the inlined struct   1         1
y      inside the inlined struct   2         2
c      AFTER the inlined struct    3         4   <-- WRONG
d      AFTER the inlined struct    4         5   <-- WRONG
```

not-found probe: -7, claiming 6 fields where there are 5

Why the conditions never meet in practice

Condition 2 is the one nothing in the tree satisfies. RPython models inheritance by inlining the base class as the first field, so every inlined sub-struct in a real rtyper-built type sits at position 0:

SUB._names = ('super', 'inst_x')
   position 0: super     INLINED STRUCT
   position 1: inst_x    plain field

Deeper hierarchies nest rather than accumulate siblings — SUBSUB(super=SUB(super=OBJECT, …), …) — so position 0 holds at every level. Instrumenting lltype.Struct.__init__ to log every GcStruct built while importing the major modules, plus a static scan of rpython/ and pypy/, found no GcStruct with an inlined sub-struct anywhere but position 0.

Two independent inlined structs as siblings would trigger it — the second one has a non-empty prefix — which is what test_index_after_two_inlined_structs covers. That shape is legal lltype; it simply has no instance today.

And nothing pointed at the gap from the other side either: rpython/jit/codewriter/test/test_heaptracker.py did not exist, so the function had no direct coverage at all.


youknowone:
failing tsets are same as c8a8b43148169f1c5b7ed76f159565e341244074

cfbolz:
yes, the CI test failures are spurious :-(. we are running more things on CI now, and it goes wrong, apparently

cfbolz:
I've merge in main, want to see whether the CI is happier now

PR #5538, opened 2026-07-30
jit: repair the stale rlist.ll_getitem() call in the codewriter builtin table

youknowone:
commit 72270f262488167a70bf4888aef9dcc42b388449 changed ll_getitem signature, but didn't update the its usage.


cfbolz:
can you explain what the effect of this problem is? does it have any effect on the real world?

youknowone:
I think this is not a real bug and they are not actually being used.

PR #5508, opened 2026-06-20
x86 jit: break cvtsi2sd false dependency with a zeroing idiom

youknowone:
CVTSI2SD preserves the destination's high 64 bits, carrying a false dependency on the register's prior value. In a loop that reuses the same xmm this serialises the int->float conversion against the previous iteration. Emit a pxor zeroing idiom (eliminated at register rename) before the conversion to break it.

Output unchanged: cvtsi2sd overwrites the low 64 bits, so zeroing the rest is semantically inert for scalar doubles.

test script
```python
def main():
    s = 0.0
    i = 0
    while i < 200000000:
        s = s + i * 0.1
        i = i + 1
    print(s)

main()
```

measurement
```
pwsh .\measure.ps1                             
run=1 name=installed-pypy2 sec=0.485363 code=0
run=1 name=built-pypy2-cvtsi2sd sec=0.146969 code=0
run=2 name=installed-pypy2 sec=0.281597 code=0
run=2 name=built-pypy2-cvtsi2sd sec=0.144933 code=0
run=3 name=installed-pypy2 sec=0.282210 code=0
run=3 name=built-pypy2-cvtsi2sd sec=0.144413 code=0
run=4 name=installed-pypy2 sec=0.282383 code=0
run=4 name=built-pypy2-cvtsi2sd sec=0.145911 code=0
run=5 name=installed-pypy2 sec=0.281303 code=0
run=5 name=built-pypy2-cvtsi2sd sec=0.144684 code=0
run=6 name=installed-pypy2 sec=0.288980 code=0
run=6 name=built-pypy2-cvtsi2sd sec=0.144944 code=0
run=7 name=installed-pypy2 sec=0.281541 code=0
run=7 name=built-pypy2-cvtsi2sd sec=0.145592 code=0
--- summary ---
installed-pypy2 avg=0.311911 median=0.282383 min=0.281303 max=0.485363
built-pypy2-cvtsi2sd avg=0.145349 median=0.145592 min=0.144413 max=0.146969
speedup_median=1.940x reduction=48.4%
```

<summary>

measurement detail

<details>

baseline version:
```
pypy2 --version
Python 2.7.18 (d3d629bf20c9, May 26 2026, 06:09:05)
[PyPy 7.3.23 with MSC v.1941 64 bit (AMD64)]
```

patched version
```
pypy\goal\pypy-c.exe --version
Python 2.7.18 (4be9667c6636, Jun 19 2026, 16:09:59)
[...]

PR #5541, opened 2026-07-30
jit: slice the kind in rewrite_op_getarrayitem so the raw_r guard can fire

youknowone:
Unify indexing to `getkind(...)[0]` to avoid manual check for every use case

### DIFF
--- a/rpython/jit/codewriter/jtransform.py
+++ b/rpython/jit/codewriter/jtransform.py
@@ -760,9 +760,9 @@ class Transformer(object):
         if op.args[0] in self.vable_array_vars:     # for virtualizables
             vars = self.vable_array_vars[op.args[0]]
             (v_base, arrayfielddescr, arraydescr) = vars
-            kind = getkind(op.result.concretetype)
+            kind = getkind(op.result.concretetype)[0]
             return [SpaceOperation('-live-', [], None),
-                    SpaceOperation('getarrayitem_vable_%s' % kind[0],
+                    SpaceOperation('getarrayitem_vable_%s' % kind,
                                    [v_base, op.args[1], arrayfielddescr,
                                     arraydescr], op.result)]
         # normal case follows
@@ -771,7 +771,7 @@ class Transformer(object):
         if immut:
             pure = '_pure'
         arraydescr = self.cpu.arraydescrof(ARRAY)
-        kind = getkind(op.result.concretetype)
+        kind = getkind(op.result.concretetype)[0]
         if ARRAY._gckind != 'gc':
             assert ARRAY._gckind == 'raw'
             if kind == 'r':
@@ -779,7 +779,7 @@ class Transformer(object):
             pure = ''   # always redetected from pyjitpl.py: we don't need
                         # a '_pure' version of getarrayitem_raw
         return SpaceOperation('getarrayitem_%s_%s%s' % (ARRAY._gckind,
-                                                        kind[0], pure),
+                                                        kind, pure),
                               [op.args[0], op.args[1], arraydescr],
                               op.result)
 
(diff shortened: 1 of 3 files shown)


rpython/jit/codewriter/jtransform.py
```diff
@@ -771,15 +771,15 @@ def rewrite_op_getarrayitem(self, op):
         if immut:
             pure = '_pure'
         arraydescr = self.cpu.arraydescrof(ARRAY)
-        kind = getkind(op.result.concretetype)
+        kind = getkind(op.result.concretetype)[0]
         if ARRAY._gckind != 'gc':
             assert ARRAY._gckind == 'raw'
             if kind == 'r':
```
youknowone, line 777: this line `kind == 'r'` was comparing `getkind(...)` to `r` instead of `getkind(...)[0]` like other usages

cfbolz:
Again, I would be interested in knowing whether there is a wider effect of this. Eg can you write a test metainterp/test/test_ajit.py that gets fixed by this change?

youknowone:
@cfbolz I didn't find this by real code, so not sure it affect to real pypy production. 

~~Added a generated test but didn't check it yet. I will check it later today~~

rpython/jit/metainterp/test/test_ajit.py
```diff
@@ -3255,6 +3255,21 @@ def f1(n):
...
+        # letting the blackhole interpreter fail to find a handler for it.
+        S = lltype.GcStruct('S', ('x', lltype.Signed))
+        A = rffi.CArray(lltype.Ptr(S))
+        def f(n):
+            a = lltype.malloc(A, 5, flavor='raw', zero=True)
+            res = 1 if a[n] else 0
+            lltype.free(a, flavor='raw')
+            return res
+        e = py.test.raises(Exception, self.interp_operations, f, [0])
+        assert 'getarrayitem_raw_r not supported' in str(e.value)
```
youknowone, line 3271: Before the patch, it raises `AttributeError: type object 'BlackholeInterpreter' has no attribute 'bhimpl_getarrayitem_raw_r'` in blackhole.py

```
  File "p4_ajit_probe.py", line 31, in <module>
    res = r.interp_operations(f, [0])
  File "rpython/jit/metainterp/test/support.py", line 273, in interp_operations
    result1 = _run_with_blackhole(self, args)
  File "rpython/jit/metainterp/test/support.py", line 113, in _run_with_blackhole
    blackholeinterpbuilder = BlackholeInterpBuilder(cw)
  File "rpython/jit/metainterp/blackhole.py", line 58, in __init__
    self.setup_insns(asm.insns)
  File "rpython/jit/metainterp/blackhole.py", line 80, in setup_insns
    all_funcs.append(self._get_method(name, argcodes))
  File "rpython/jit/metainterp/blackhole.py", line 238, in _get_method
    unboundmethod = getattr(BlackholeInterpreter, 'bhimpl_' + name).im_func
AttributeError: type object 'BlackholeInterpreter' has no attribute 'bhimpl_getarrayitem_raw_r'
```

youknowone:
For more details.

I am working for https://github.com/youknowone/pyre, and because it requires agents to read PyPy code a lot, it occasionally find PyPy bugs and suspicious code smells. Sometimes it is a production glitch (#5508) and sometimes potential risk like this. I actually got more reports about smells but didn't submit them because I wasn't sure they are real bugs.
To be honest, I don't understand pypy codebase well to understand the reports enough and patch myself. I try to confirm they are real issues. I am not sure the patch is correct, but usually sure the report is real.
@cfbolz if you don't like patches like this, not real bugs observable in pypy, please feel free to tell me. I can stop them and submit only real production bug fixes. Otherwise I will try to append enough test and explanation, but unlike the real bugs, tests and explanation about them will be mostly generated because I don't understand them that much.

cfbolz:
No, I like your patches! They seem to find real corner cases. I just want to 1) think about them for a few days because I'm worried about unintended consequences and 2) make sure that the right kind of test is also added.

PR #5548, opened 2026-07-31
jit: add the missing opimpl_record_known_result_r_ir_v

youknowone:
This is a blackhole interpreter bug, but I don't really well understand the detail. This is the last patch of the series this turn. 

AI-generated description below.

----

`jit.record_known_result()` works when the elidable function returns an int and
crashes when it returns a GC pointer. Half of the opcode is missing from the
metainterp.

`rewrite_op_jit_record_known_result()` picks the spelling from the result kind:

```python
if getkind(op.args[0].concretetype) == "int":
    opname = "record_known_result_i"
else:
    assert getkind(op.args[0].concretetype) == "ref"
    opname = "record_known_result_r"
```

`blackhole.py` implements both `bhimpl_record_known_result_i_ir_v` and
`bhimpl_record_known_result_r_ir_v`. `pyjitpl.py` implements only the `_i`
spelling, and `setup_insns()` resolves every emitted opcode name through
`getattr(MIFrame, 'opimpl_' + name)`, so the `_r` one has nothing to resolve to.

### Reproducer

Twenty lines are enough:

```python
class W(object):
    def __init__(self, x):
        self.x = x

CACHE = [W(0), W(10), W(20), W(30)]

@elidable
def get_ref(i):
    return CACHE[i]                      # returns a GC pointer

@elidable
def get_int(i):
    return CACHE[i].x

def main_ref(i):
    w = get_ref(i)
    record_known_result(w, get_ref, i)   # -> record_known_result_r
    return w.x

def main_int(i):
    n = get_int(i)
    record_known_result(n, get_int, i)   # -> record_known_result_i
    return n
```

Run through `LLJitMixin.interp_operations(fn, [2])`:

```
main_int  (record_known_result_i)   returned 20
main_ref  (record_known_result_r)   AttributeError: type object 'MIFrame' has no
                                    attribute 'opimpl_record_known_result_r_ir_v'
```

Nothing refuses the shape on the way there. The codewriter does not treat a ref
result as unsupported — it asserts that a non-int result kind *is* `ref` and
emits the opcode for it — so the failure is a missing attribute named after an
internal opcode, at codewriter setup time, rather than anything pointing back at
the `record_known_result()` call.

### The fix

The `_i` body does not look at the result's kind: it builds the boxes and
records `rop.RECORD_KNOWN_RESULT`. So the two spellings can share one
implementation, the way they already do in `blackhole.py`, where both bodies are
empty. This patch adds the alias.

### Reachability

Nothing in the tree reaches this today, and I would rather say so up front. The
only caller, `pypy/objspace/std/unicodeobject.py`, passes an int result:

```python
jit.record_known_result(self._length, rutf8._check_utf8, self._utf8, True, 0, -1)
```

What makes it worth filling in rather than guarding is that `record_known_result`
is a documented public API — it is announced in `doc/release-v7.3.10.rst` — and
an elidable function returning an object is an ordinary thing to write. The
correct behaviour is not an open question either: three of the four layers
already handle a ref result.

### Test

`test_record_known_result` in `rlib/test/test_jit.py` stops at `t.rtype()` and
never runs the codewriter, so neither spelling is covered there; that is why the
gap survived. The added test asserts that every
`bhimpl_record_known_result_*` in `blackhole.py` has a matching `opimpl_` in
`pyjitpl.py`. Without the change it fails with

```
AssertionError: blackhole.py implements bhimpl_record_known_result_r_ir_v
but pyjitpl.py has no opimpl_record_known_result_r_ir_v
```

`metainterp/test/test_pyjitpl.py` goes from 2 passed / 1 skipped to 3 passed /
1 skipped, i.e. exactly the new test.

### DIFF
--- a/rpython/jit/metainterp/pyjitpl.py
+++ b/rpython/jit/metainterp/pyjitpl.py
@@ -418,6 +418,8 @@ class MIFrame(object):
         # an operation
         self.metainterp._record_helper_varargs(rop.RECORD_KNOWN_RESULT, None, calldescr, allboxes)
 
+    opimpl_record_known_result_r_ir_v = opimpl_record_known_result_i_ir_v
+
     @arguments("box", "box")
     def opimpl_record_exact_value_r(self, box, const_box):
         return self._record_exact_value(rop.RECORD_EXACT_VALUE_R, box, const_box)
--- a/rpython/jit/metainterp/test/test_ajit.py
+++ b/rpython/jit/metainterp/test/test_ajit.py
@@ -4149,6 +4149,34 @@ class BaseLLtypeTests(BasicTests):
         assert res == main(10)
         self.check_resops(call_i=2)  # two calls to f, both get removed by the backend
 
+    def test_record_known_result_ref(self):
+        # the elidable function returns a GC pointer, so this goes through
+        # record_known_result_r rather than record_known_result_i
+        class W(object):
+            def __init__(self, x):
+                self.x = x
+        cache = [W(i) for i in range(20)]
+
+        @elidable
+        def f(x):
+            return cache[x]
+
+        def call_f(x):
+            w = f(x)
+            record_known_result(w, f, x)
+            return w.x
+
+        myjitdriver = JitDriver(greens=[], reds=['x', 'res'])
+        def main(x):
+            res = 0
+            while x > 0:
+                myjitdriver.jit_merge_point(x=x, res=res)
+                res += call_f(x)
+                x -= 1
+            return res
+        res = self.meta_interp(main, [10], backendopt=True)
+        assert res == main(10)
+
 
     def test_record_exact_value(self):
         class A(object):


rpython/jit/metainterp/pyjitpl.py
```diff
@@ -418,6 +418,12 @@ def opimpl_record_known_result_i_ir_v(self, resbox, funcbox, argboxes,
         # an operation
         self.metainterp._record_helper_varargs(rop.RECORD_KNOWN_RESULT, None, calldescr, allboxes)
 
+    # jtransform.py:303-307 also emits 'record_known_result_r' when the
```
cfbolz: Remove this comment block

cfbolz:
This looks like an oversight, thanks for finding it. We should still add an integration test that uses this kind of record known result call.

cfbolz:
looks reasonable, thanks

PR #5552, opened 2026-08-03
jit: thread handling_ovf into generate_last_exc() so a re-raised OverflowError survives

youknowone:
generated report below. I asked to scan every re-raise patterns and currently no code triggers this

----

An `OverflowError` from `ovfcheck()` that is re-raised out of its handler and
caught again in the same function raises `AttributeError` through the JIT.

## Reproducer

```python
from rpython.rlib.rarithmetic import ovfcheck

def f(n):
    try:
        try:
            return ovfcheck(n * n)
        except OverflowError:
            raise              # caught again just below
    except OverflowError:
        return -1
```

## How to run it

```python
from rpython.jit.metainterp.test.support import LLJitMixin

class Runner(LLJitMixin):
    pass

print f(4000000000)                                    # plain Python
print Runner().interp_operations(f, [4000000000])      # through the JIT
```

## Expected vs actual

Expected, and what plain Python and llinterp both give: `-1`.

Actual, through the JIT:

```
  File "rpython/jit/metainterp/blackhole.py", line 989, in bhimpl_last_exception
    real_instance = self.exception_last_value
AttributeError: 'BlackholeInterpreter' object has no attribute 'exception_last_value'
```

Blackhole trace, last two lines:

```
bh: int_mul_jump_if_ovf [11, 4000000000, 4000000000] -> (0, 11)
bh: last_exception [<BHInterp <JitCode 'f'>>] -> AttributeError!
```

## Emitted jitcode

Before:

```
 3  int_mul_jump_if_ovf L1, %i0, %i0 -> %i0
 9  int_return %i0
11  last_exception -> %i1
13  last_exc_value -> %r0
15  inline_call_ir_i <JitCode 'll_issubclass_const__object_vtablePtr_Signed_Signed'>, ...
```

After:

```
 3  int_mul_jump_if_ovf L1, %i0, %i0 -> %i0
 9  int_return %i0
11  int_copy $<* struct object_vtable> -> %i1
14  ref_copy $<* struct object> -> %r0
33  int_return $-1
38  raise %r0
```

## Which shapes are affected

`interp_operations(fn, [4000000000])`:

| shape | before | after |
|---|---|---|
| `try: ovfcheck(n*n)` / `except OverflowError: return -1` | `-1` | `-1` |
| the same, binding `except OverflowError as e` | `-1` | `-1` |
| nested, inner handler is a bare `raise` | **`AttributeError`** | `-1` |
| nested, inner handler does work, then `raise` | **`AttributeError`** | `-1` |
| `raise` with no outer handler in this function | `LLException 'OverflowError'` | unchanged |

## Tests

* `metainterp/test/test_ajit.py::test_ovf_reraise_caught_again_in_the_same_function`
  — end-to-end, covering both the bare re-raise and the does-work shape. Without
  the patch: `AttributeError: 'BlackholeInterpreter' object has no attribute
  'exception_last_value'`.
* `codewriter/test/test_flatten.py::test_ovfcheck_reraise_caught_again_in_the_same_function`
  — asserts `int_add_jump_if_ovf` is emitted and `last_exception` /
  `last_exc_value` are not. Without the patch the emitted assembler still
  contains `last_exception -> %i3`.

`encoding_test()` in `test_flatten.py` now returns the formatted assembler; it
previously returned `None`.

### DIFF
--- a/rpython/jit/codewriter/flatten.py
+++ b/rpython/jit/codewriter/flatten.py
@@ -151,7 +151,7 @@ class GraphFlattener(object):
             and link.last_exc_value not in link.args):
             self.make_return(link.args)     # optimization only
             return
-        self.insert_renamings(link)
+        self.insert_renamings(link, handling_ovf)
         self.make_bytecode_block(link.target, handling_ovf)
 
     def make_exception_link(self, link, handling_ovf):
@@ -303,7 +303,7 @@ class GraphFlattener(object):
                 self.emitline('-live-')
                 self.make_link(switch, handling_ovf)
 
-    def insert_renamings(self, link):
+    def insert_renamings(self, link, handling_ovf):
         renamings = {}
         lst = [(self.getcolor(v), self.getcolor(link.target.inputargs[i]))
                for i, v in enumerate(link.args)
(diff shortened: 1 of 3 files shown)


cfbolz:
looks good, merging

PR #5554, opened 2026-08-04
Avoid deprecated sysconfig check_home arguments

stanbot8:
CPython fixed the same Python 3.11 warning in [python/cpython#98741](https://github.com/python/cpython/issues/98741). [PyPy 3.12](https://github.com/pypy/pypy/commit/9429f81e6550) already contains the corrected calls. Its py3.11 branch retained the deprecated arguments.

PyPy emits a `DeprecationWarning` when distutils or venv checks whether it is running from a source build.

Call `sysconfig.is_python_build` without its deprecated `check_home` argument in distutils and venv. Add regression coverage that promotes the `check_home` deprecation to an error in both runtime paths.


mattip:
Thanks

PR #5558, opened 2026-08-10
Fix `bytes()`, `bytearray()` and `codecs` accepting `encoding=None`

jseop-lim:
## Description

`bytes()`, `bytearray()`, `codecs.encode()` and `codecs.decode()` accept an explicit `None` for `encoding` or `errors` and carry on with the default, where CPython raises `TypeError`. All four declare `@unwrap_spec(encoding='text_or_none', errors='text_or_none')`, and `space.text_or_none_w` maps an app-level `None` onto the RPython `None` an omitted argument already carries, so nothing downstream can tell `bytearray(0, None)` from `bytearray(0)`. CPython types these parameters as clinic `str` and decides presence from `fastargs[1] != NULL` before checking the value, so a supplied `None` reaches [`_PyArg_BadArgument`](https://github.com/python/cpython/blob/v3.11.15/Objects/clinic/bytearrayobject.c.h#L36-L40). `str()` already gets this right in PyPy, since it takes its two arguments wrapped and unwraps them in `get_encoding_and_errors`.

#### AS-IS
```
>>>> bytearray(0, None)
bytearray(b'')
>>>> bytes((), None)
b''
>>>> codecs.encode('a', 'utf-8', None)
b'a'
>>>> codecs.decode(b'a', 'utf-8', None)
'a'
```

#### TO-BE
```
>>> bytearray(0, None)
TypeError: bytearray() argument 'encoding' must be str, not None
>>> bytes((), None)
TypeError: bytes() argument 'encoding' must be str, not None
>>> codecs.encode('a', 'utf-8', None)
TypeError: encode() argument 'errors' must be str, not None
>>> codecs.decode(b'a', 'utf-8', None)
TypeError: decode() argument 'errors' must be str, not None
```
(matching CPython 3.11.15)

For the codecs the `None` is not merely a wrong type: it is read as `'strict'`, so code that passed `errors='ignore'` through a wrapper gets a `UnicodeEncodeError` that CPython never reaches. `bytes('abc', None)` and `bytearray('abc', None)` also change, from `string argument without an encoding` to the message above.

## Changes

- Add `space.text_arg_w()` next to `text_or_none_w()`. It unwraps a `str` and reports a failure the way the Argument Clinic converter does. It takes only wrapped values, so the caller still decides what an omitted argument means.
- Take `encoding` and `errors` wrapped in `bytes.__new__`, `bytearray.__init__`, `codecs.encode` and `codecs.decode`, and route them through it. `text_or_none` itself is unchanged: the AST and typedef layers rely on its current meaning.
- Add tests for the four constructors and functions.

### DIFF
--- a/pypy/interpreter/baseobjspace.py
+++ b/pypy/interpreter/baseobjspace.py
@@ -1822,6 +1822,22 @@ class ObjSpace(object):
     def text0_or_none_w(self, w_obj):
         return None if self.is_none(w_obj) else self.text0_w(w_obj)
 
+    def text_arg_w(self, w_obj, funcname, argname):
+        """Like text_w(), but reporting a failure the way the str converter
+        of CPython's Argument Clinic does.  Use it where an app-level None
+        must not be folded onto an omitted argument, which is what
+        text_or_none_w() above does."""
+        if self.is_none(w_obj):
+            # _PyArg_BadArgument() spells this type "None", not "NoneType"
+            raise oefmt(self.w_TypeError,
+                        "%s() argument '%s' must be str, not None",
+                        funcname, argname)
+        if not self.isinstance_w(w_obj, self.w_unicode):
+            raise oefmt(self.w_TypeError,
+                        "%s() argument '%s' must be str, not %T",
+                        funcname, argname, w_obj)
+        return self.text_w(w_obj)
+
     @specialize.argtype(1)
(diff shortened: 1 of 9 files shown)


mattip:
Looks good to me. I found another place that needs this as well, so I added it. Thanks!

PR #5559, opened 2026-08-10
Fix `int()` raising `TypeError` instead of `ValueError` for a bad base

jseop-lim:
## Description

`int(x, base)` with a non-string `x` and an out-of-range `base` raises `TypeError`, where CPython raises `ValueError`. `_new_baseint` converts `base` with `space.getindex_w` but leaves the range check to the number parser in `rpython/rlib/rstring.py`, which is only reachable once `x` has been accepted as `str`, `bytes`, or `bytearray`. CPython's [`long_new_impl`](https://github.com/python/cpython/blob/v3.11.15/Objects/longobject.c#L5380-L5390) checks the range before it dispatches on the type of `x`.

#### AS-IS
```
>>>> int(None, 1)
TypeError: int() can't convert non-string with explicit base
```

#### TO-BE
```
>>> int(None, 1)
ValueError: int() base must be >= 2 and <= 36, or 0
```
(matching CPython 3.11.15)

## Changes

- Range-check `base` in `_new_baseint` right after `space.getindex_w`, ahead of the `str` / `bytes` / non-string dispatch. The string path already produces this exact message through `rpython/rlib/rstring.py`, so the check moves rather than a new message being introduced.
- Add a test that an out-of-range base wins over the non-string `TypeError`.

### DIFF
--- a/pypy/objspace/std/intobject.py
+++ b/pypy/objspace/std/intobject.py
@@ -1075,7 +1075,14 @@ def _new_baseint(space, w_value, w_base=None):
         except OperationError as e:
             if not e.match(space, space.w_OverflowError):
                 raise
-            base = 37 # this raises the right error in string_to_bigint()
+            base = 37 # out of range, so the check below raises
+
+        # CPython validates the base range in long_new_impl(), before it
+        # dispatches on the type of the value.  The message is the one
+        # NumberStringParser builds for the string path.
+        if base != 0 and (base < 2 or base > 36):
+            raise oefmt(space.w_ValueError,
+                        "int() base must be >= 2 and <= 36, or 0")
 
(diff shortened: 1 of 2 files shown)


mattip:
Does this have a cpython (stdlib) test that is failing?

mattip:
Thanks. To answer my own question: no, there is no failing stdlib test for this, it is entirely an untested edge case and unreported in an issue in PyPy.

jseop-lim:
> Does this have a cpython (stdlib) test that is failing?

@mattip Sorry I missed this. I should have answered your question myself.

> Thanks. To answer my own question: no, there is no failing stdlib test for this, it is entirely an untested edge case and unreported in an issue in PyPy.

Thanks for looking into it and for accepting the PR!

Issue #5275, opened 2025-05-19
Py3.12: PEP 669: Low impact monitoring for CPython


cfbolz:
This is an optional CPython feature, low priority

https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep669

mattip:
I asked claude to make a plan, [this is what we came up with](https://gist.github.com/mattip/fefb67a4b65414437c2d4d5421e17896) after a few iterations. In the mean time, I skipped the lib-python tests on 3.12 with a sys.monitoring attribute check.

mattip:
Closing as finished, #5569 and subsequent fixes.

PR #5569, opened 2026-08-19
Implement PEP 669 low impact sys.monitoring

mattip:
It turns out the idea of a JIT friendly `VersionTag` singleton was enough to unlock a performant version of sys.monitoring for PyPy. There is no impact when the monitoring is off, and after being disabled once a few events callbacks have been called, performance returns to the no impact level. So this PR became threading the `VersionTag` conditional through all the instructions that are monitored, and working out details like what happens if monitoring is toggled mid-loop.

Benchmarks show parity with pypy3.12-HEAD, plus-minus typical variations.

This script was used to test the post-disable performance

```
import sys
import time

N = 50_000_000


def hot_loop(n):
    i = 0
    total = 0
    while i < n:
        total += i
        i += 1
    return total


t0 = time.time()
hot_loop(N)
print('unmonitored:            %.4fs' % (time.time() - t0))

E = sys.monitoring.events
sys.monitoring.use_tool_id(3, "bench")

hits = [0]


def jump_cb(code, offset, dest):
    hits[0] += 1
    return sys.monitoring.DISABLE


sys.monitoring.register_callback(3, E.JUMP, jump_cb)
sys.monitoring.set_local_events(3, hot_loop.__code__, E.JUMP)

t0 = time.time()
hot_loop(N)
print('JUMP DISABLE-after-hit: %.4fs (callback fired %d time(s))' % (time.time() - t0, hits[0]))

sys.monitoring.set_local_events(3, hot_loop.__code__, 0)
sys.monitoring.free_tool_id(3)
```

which on my machine prints
```
unmonitored:            0.0385s
JUMP DISABLE-after-hit: 0.0353s (callback fired 2 time(s))
```

AI was used heavily to develop this, but I monitored and steered the work closely.


mattip:
Closes #5275

mattip:
This also has implications for the built-in _lsprof module which could be refactored to use sys.monitoring, fixing some test failures in `test_cprofile`.

mattip:
We discussed this and decided to merge it.

PR #5573, opened 2026-08-24
rather than allocate at each enter_code call, use a freelist

mattip:
Using perf, I noticed an allocation every time we enter a frame. Rather than allocate, use a freelist (`rthread.ThreadLocalField(PVMPROFSTACK, "vmprof_tl_freelist")`) which speeds up `sympy_integrate` and `sqlalchemy_imperative` (call heavy benchmarks that the JIT cannot help much) by 10%

### DIFF
--- a/pypy/interpreter/generator.py
+++ b/pypy/interpreter/generator.py
@@ -198,7 +198,7 @@ return next yielded value or raise StopIteration."""
                     jitdriver.jit_merge_point(self=self, frame=frame,
                                               results=results, pycode=pycode)
                     try:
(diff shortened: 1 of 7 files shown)


mattip:
Also inlined a frame.

mattip:
The speedup on main is significant, the speedup on py3.11 less so - only about 3-4%.

mattip:
The single-benchmark speedup did not really achieve much. The JIT, even when not very successful in speeding up the benchmark, masks the effect of this change. Even the non-jit sympy_integrate benchmark changed by only about 5%.

PR #5576, opened 2026-09-01
Pyrepl syntax highlighting

cfbolz:
This ports CPython 3.14's interactive syntax highlighting to PyPy's Python 3.12 pyrepl and brings the closely related renderer and input behavior in line with upstream.

image


StanFromIreland:
Yay! Thanks for doing this.

PR #5578, opened 2026-09-06
enable cp312-abi3 wheel imports

mattip:
This should complete the idea from #3397 to allow pypy to import abi3 wheels if they are built for python3.12 and up.

There is a hack to monkey patch packaging (used by pip and others) to allow pypy to download and install cp312-abi3 wheels. That should be upstreamed to the official channels once we are sure this works.

Still TODO: Windows needs a python3.dll reflector like CPython. 

With this branch, I seem to be able to install onnx (but its numpy dependency requires a build-from-source) which is based on nanobind. I can also install and run pyzmq, which is based on cython. It imports, tries to get the cffi backend, fails since the abi3 wheel does not ship that backend, and loads the cython based one instead. Creating a socket PAIR round-trip works.


auvipy (approved):
looks good

PR #5579, opened 2026-09-08
backport PyObject prefix config option

mattip:
Backport py3.12 rpython changes to main. Adds a new option to store the `ob_pypy_link` field of PyObject in a prefix. The option is True for py3.12+, and False for existing py2.7 (main) and py3.11 branches. Also add documentation for the option and how it all works.


mattip:
Done differently. Closing

